Issue #18, opened 2017-09-26
UI scaling on high-dpi screen


dvicini:
Hi,
The UI does currently not seem to scale with higher DPI screens. I am using Arch Linux (KDE) and a 4K monitor. The KDE user interface scales nicely, but the tev interface becomes tiny.

I am not sure whether the underlying issue here is a nanogui problem, so feel free to close the issue if you think it should be fixed in nanogui itself. ;)
Delio

Tom94:
Hi Delio,

this indeed is a problem at nanogui's end. Currently, on linux, it only obtains the pixel ratio via gsettings, so if you are not using gnome it probably won't work. Since tev is already using a fork of nanogui with some similar fixes implemented, I'll have a look at how easy scaling is to add for KDE. In case you want to look into it yourself, given that you are able to replicate the problem, [here](https://github.com/Tom94/nanogui/blob/master/src/screen.cpp#L68) is the file and function in question.

Cheers,
Thomas

dvicini:
For reference, I commented on this in the nanogui main repository
https://github.com/wjakob/nanogui/issues/202

Tom94:
Thanks for the help! I will test this locally on my machine and will merge afterwards. :)

PR #23, opened 2017-09-29
Support hi-dpi scaling in KDE (kudos to Delio Vicini)

Tom94:
See https://github.com/wjakob/nanogui/pull/265

Since __tev__ already supports fractional pixel ratios, Delio's change can be merged as-is once tested.

Fixes https://github.com/Tom94/tev/issues/18


Tom94:
Confirmed working.

PR #36, opened 2018-03-06
Add package link for archlinux

oktomus:
[https://aur.archlinux.org/packages/tev/](https://aur.archlinux.org/packages/tev/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,13 @@ $ tev -h
 
 ## Obtaining tev
 
-Pre-built binaries for Windows (32-bit and 64-bit) and macOS (64-bit) are available on the [releases page](https://github.com/Tom94/tev/releases). For other operating systems __tev__ has to be built from source as of now (next section).
+### Mac OS X / Windows
+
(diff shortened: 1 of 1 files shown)


Tom94:
Hi oktomus,

thanks for putting tev on the AUR. However, I am concerned that some dependencies might be missing from the PKGBUILD, namely X, OpenGL, and zlib. Are these perhaps implicitly handled somehow? Unfortunately I do not have a clean Arch install lying around to check that the installation works out-of-the-box.

Cheers,
Tom94

oktomus:
Hey @Tom94, yes they are :) 
I tested on my installation but I will test it on a clean installation soon. 
- zlib is in the core
- X is installed as soon as you have a graphic environment
Except opengl, that I will add now !

oktomus:
Tested on a clean installation, building failling because it hasn't rights to copy in /usr. I will try to fix it, still new with package building :)

oktomus:
Package fixed, builds and installs like a charm :)

Tom94:
Perfect, this looks a lot better. Thank you very much! :)

PR #60, opened 2018-12-04
Fix for parsing PFM header

gargll:
Previously, the PFM parser used seekg(1) to move past the newline
at the end of the header. For files created using CR+LF end-of-line
marker, seekg(2) would be needed. Calling std::getline() instead handles
that transparently.


Tom94:
Thanks a lot! I wish the local variable wasn't needed, but I could not find a nicer approach than yours. I'll push a new release so the binaries are up-to-date.

PR #66, opened 2019-03-16
Adds the missing supported PGM and PPM file extensions

### LINKED ISSUE #65: Only shows .pnm as supported extension, not .ppm, .pgm or .pbm
Hi,

Tev supports the PNM format which includes PPM, PGM and PBM (according to [wikipedia](https://en.wikipedia.org/wiki/Netpbm_format)). However, in the "open" dialog of tev, only `.pnm` is listed as supported, leading to files with the other three extensions not shown in the dialog. 

tev-support-ext

Should be a simple fix, shouldn't it? :)

Tom94:
Note, that only the binary versions of the above formats are supported (a limitation incurred by using STB). Furthermore, the PBM and PAM formats (also part of PNM) are not supported at all.

Fixes #65

### DIFF
--- a/src/ImageViewer.cpp
+++ b/src/ImageViewer.cpp
@@ -1238,19 +1238,21 @@ void ImageViewer::openImageDialog() {
     vector<string> paths = file_dialog(
     {
         // HDR formats
-        {"exr",  "OpenEXR Image"},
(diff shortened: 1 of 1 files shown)


MKimiSH:
Thanks for the quick fix :p

PR #67, opened 2019-04-10
Keyboard shortcut to close all images

merlinND:
I found this useful when working with image sequences with hundreds of frames.
I tried to respect the coding style, but let me know if you would like me to adjust something!

---

- [x] `removeAllImages` function and key binding
- [x] Entry in the Help panel

Tested on Ubuntu 18.10.

### DIFF
--- a/src/ImageViewer.cpp
+++ b/src/ImageViewer.cpp
@@ -656,7 +656,9 @@ bool ImageViewer::keyboardEvent(int key, int scancode, int action, int modifiers
             }
         }
 
-        if (key == GLFW_KEY_W && modifiers & SYSTEM_COMMAND_MOD) {
+        if (key == GLFW_KEY_W && modifiers & SYSTEM_COMMAND_MOD && modifiers & GLFW_MOD_SHIFT) {
+            removeAllImages();
+        } else if (key == GLFW_KEY_W && modifiers & SYSTEM_COMMAND_MOD) {
             removeImage(mCurrentImage);
         } else if (key == GLFW_KEY_UP || key == GLFW_KEY_W || key == GLFW_KEY_PAGE_UP) {
             if (modifiers & GLFW_MOD_SHIFT) {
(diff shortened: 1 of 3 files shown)


src/ImageViewer.cpp
```diff
@@ -857,6 +859,27 @@ void ImageViewer::removeImage(shared_ptr<Image> image) {
     }
 }
 
+void ImageViewer::removeAllImages() {
+    if (mImages.empty())
+        return;
+
+    // Reset all focus as a workaround a crash caused by nanogui.
+    // TODO: Remove once a fix exists.
```
Tom94: Interesting... is this crash connected with the one that happens sometimes when closing many images quickly in a row? I've been trying to hunt that one down for a long time.

src/ImageViewer.cpp
```diff
@@ -857,6 +859,27 @@ void ImageViewer::removeImage(shared_ptr<Image> image) {
...
+    requestFocus();
+
+    for (long i = mImages.size() - 1; i >= 0; --i) {
```
Tom94: This would be more correct as `for (size_t i = mImages.size(); i > 0; --i) {` with a subtraction of `1` from `i` within the loop body.

src/ImageViewer.cpp
merlinND: I actually copy-pasted this comment and the `requestFocus()` call from `removeImage` :stuck_out_tongue: 
I don't experience a crash when excluding it, but I thought I shouldn't risk it in case it's a platform-specific issue, and included it anyway.

merlinND:
Thanks for the detailed review! I've applied the comments.

I had also thought about the button situation and I agree that adding another one is not the best idea.
We could take the shift modifier into account when clicking the Close button. If included in the tooltip + the help panel, it makes the feature discoverable-enough I would say.

src/ImageViewer.cpp
Tom94: Whoopsie ^_^;

Tom94:
Looks good. Thanks again for implementing this. :)
I've gone ahead and added the shift modifier to the button and its tooltip; will merge once CI is through.

PR #72, opened 2019-05-18
Build using VS2019 on Windows

Tom94:


Tom94:
Obsolete after switch to GitHub actions

PR #75, opened 2019-08-27
Update stb_image_write.h

jammm:
Uses version 1.13. Not tested yet though.


Tom94:
Thanks a lot for the PR!

Is there any particular feature of the newer STB you need in tev? I am hesitant to spend the time testing as long as there isn't really any benefit to updating (the changelog doesn't indicate anything new that'd be relevant).

jammm:
Actually I just noticed that it's an older version so thought a PR might be worthwhile :P
If there's no benefit to updating, then let's close this for now.

PR #77, opened 2019-09-27
Keep Target Position During Drag-To-Zoom Constant

reiner-dolp:
Otherwise, zooming into an area of interest is nearly impossible.

### DIFF
--- a/src/ImageViewer.cpp
+++ b/src/ImageViewer.cpp
@@ -452,9 +452,11 @@ bool ImageViewer::mouseButtonEvent(const Vector2i &p, int button, bool down, int
     if (down) {
         if (canDragSidebarFrom(p)) {
             mIsDraggingSidebar = true;
+            mDraggingStartPosition = p.cast<float>();
             return true;
         } else if (mImageCanvas->contains(p)) {
             mIsDraggingImage = true;
+            mDraggingStartPosition = p.cast<float>();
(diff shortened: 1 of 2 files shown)


src/ImageViewer.cpp
```diff
@@ -452,9 +452,11 @@ bool ImageViewer::mouseButtonEvent(const Vector2i &p, int button, bool down, int
     if (down) {
         if (canDragSidebarFrom(p)) {
             mIsDraggingSidebar = true;
+            mDraggingStartPosition = p.cast<float>();
```
Tom94: Thank you very much for your contribution!

Is there a specific reason for recording the start position when resizing the sidebar? Removing this line does not change anything as far as I can tell. Let me know and I'll take care of merging with or without accordingly. :)

reiner-dolp: The line is currently not necessary. It's just there to ensure that the variable always has the valid starting pos during dragging. Otherwise other programmers could be in for a surprise given the variable name.

We could rename the property to `mStartingPositionOfCanvasDrag` or something similar and remove the line.

But thats just bikeshedding, I'm ok with both variants.

Tom94: Fair point, I didn't really give potential future uses of the variable much thought. Thanks again!

PR #104, opened 2020-06-09
Switch to TCP for IPC

mmp:
(Only tested on OSX, but everything seems to work.)

I made a rudimentary attempt to get the Windows stuff wired up but am sure it's broken.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,6 +25,3 @@
 [submodule "dependencies/clip"]
 	path = dependencies/clip
 	url = https://github.com/Tom94/clip
-[submodule "dependencies/enet"]
-	path = dependencies/enet
-	url = https://github.com/lsalzman/enet
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,7 +85,7 @@ elseif (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID MATCHES "
     endif()
 endif()
 
-set(TEV_LIBS clip IlmImf nanogui enet ${NANOGUI_EXTRA_LIBS})
+set(TEV_LIBS clip IlmImf nanogui ${NANOGUI_EXTRA_LIBS})
 if (MSVC)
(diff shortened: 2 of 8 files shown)


Tom94:
Thanks a lot for the PR, Matt. Works like a charm!

I quickly fixed up formatting, finished Windows support, and added a bunch of extra error handling for additional corner cases.

Could you let me know if things still work on your end before I press the green button? (I was able to test macOS and Windows, but not linux beyond CI.)

mmp:
Great! (And sorry for leaving the Windows stuff on your plate...)

Unfortunately I haven't been able to text on Linux: my linux system is
head-less and I've thus far been unsuccessful at getting remote X-windows
stuff to work with it.

My only comment is regarding 2c84063923ab157dc: if I revert that, I can
successfully render on Linux and display remotely on OSX, which is nice to
be able to do. (And with that in place, I'm not able to.) Admittedly I
don't know enough about networking to understand the underlying concerns,
but it would be nice to be able to do that, at least as an option..

Tom94:
Hi Matt,

no problem at all! :)

You should already be able to remote-render as long as you pass the correct (non-local) hostname to tev for binding: i.e. start the tev server using

`tev --hostname your-local-network-ip:14158`

Cheers,
Thomas
On 2020-06-09 15:41:47, Matt Pharr <notifications@github.com> wrote:
Great! (And sorry for leaving the Windows stuff on your plate...)

Unfortunately I haven't been able to text on Linux: my linux system is
head-less and I've thus far been unsuccessful at getting remote X-windows
stuff to work with it.

My only comment is regarding 2c84063923ab157dc: if I revert that, I can
successfully render on Linux and display remotely on OSX, which is nice to
be able to do. (And with that in place, I'm not able to.) Admittedly I
don't know enough about networking to understand the underlying concerns,
but it would be nice to be able to do that, at least as an option..

mmp:
Ah, yes--confirmed: `--hostname` with the actual IP address works in that case. Thanks!

Issue #109, opened 2020-11-26
PF4 (PFM but with RGBA)


Wouterdek:
tev already supports PFM, which is very useful for quickly/easily writing out HDR images.
However, it is currently limited to monochrome or RGB. To add RGBA support, an extension exists called PF4 which adds an extra alpha channel.
You can find more details here: [https://www.cse.cuhk.edu.hk/~ttwong/data/hdrfire/hdrfire.html](https://www.cse.cuhk.edu.hk/~ttwong/data/hdrfire/hdrfire.html)

The only change compared to PFM is the magic keyword in the header, and the extra 4 bytes per pixel, so adding support should be fairly easy.

Thank you for creating tev!

PR #110, opened 2020-11-27
Add support for PF4 to PFM loader

Wouterdek:
This PR adds support for the PF4 extension to the PFM format, as described in issue #109.

Two changes were made:
 - Modify the parsing of the magic word in the header to detect PF4 (RGBA) images
 - Replace `getline` with a loop that works for both `\r` and `\n`.

Reference images from both the [PFM website](http://www.pauldebevec.com/Research/HDR/PFM/) and the [PF4 website](https://www.cse.cuhk.edu.hk/~ttwong/data/hdrfire/hdrfire.html) now work correctly.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: 
+on:
   push:
     branches:
       - master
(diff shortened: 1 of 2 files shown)


Tom94:
Thanks a bunch and apologies for the delay. :)

PR #118, opened 2021-03-13
Add DDS support (windows only)

cek:
This PR adds support for DDS images, including BC1-BC7 compressed formats.

In this PR, DDS support is only included in the windows build. The new `DirectXTex` dependency, which is used to do most of the heavy decompression and conversion lifting, did recently add some flavor of linux support, but I don't currently have easy access to a linux machine to test it.

Some notes on the PR:

In theory, it is required that COM be initialized (and then uninitialized) by a thread prior to it making any `DirectXTex` calls. In practice, I don't believe any of the codepaths that this change exercise requires COM, but I've played by the rules just to be future-safe. The current implementation is slightly overkill, in that it initializes and uninitializes COM for each call to `DdsImageLoader::load()`, rather than once at app start and teardown. However, there's no significant overhead to the current implementation, and it allows arbitrary threads to call ::load().

As noted in the comments, the dds loader assumes that any 3- or 4-channel, non-floating point DDS image is stored in sRGB space, and converts it to linear at load time. This isn't always the right guess in practice, but it seems to be the majority of the time. (In a sane world, we'd be able to assume that only *_SRGB format images were in sRGB space. But alas.)


cek:
I'll have a look at the build failure. (Windows SDK version issue, maybe...?)

cek:
Ha, this is a good one.

In version 1809 of the Windows 10 SDK, which is being used on the build machine, `wtypes.h` attempts:

`#define _VARIANT_BOOL /##/`

The compiler issues a warning re: the resulting "//" not being a valid token -- but it is a valid start of a comment (!), which apparently doesn't play well with `oaidl.h`'s presumed use of VARIANT_BOOL. 

There's a short thread on the subject here: https://stackoverflow.com/questions/9166235/vc-allows-as-the-value-of-a-macro-how-to-deal-with-it-in-eclipse-mingw

I'm not sure how to fix this issue using version 1809 of the windows SDK, given that `DirectXTex` doesn't directly include `oaidl.h` It's not installed on the machine I used to develop the PR, in fact; it's being included through some other SDK header.

Updating to a more recent version than 1809 (support for which has been discontinued) would fix the issue, I think, but I don't know if that's feasible.

Tom94:
Thanks a bunch for this super clean PR!

And also, thank you for troubleshooting the failing CI. This is really one of those "Holy Shit C++" problems.

I'll look into upgrading the CI machine to an SDK>1809 and then merge asap. :)

Tom94:
Finally got it. Thanks again! :)

PR #130, opened 2021-07-25
Add support for HDR/Apple EDR displays as well as 10 bit color

Tom94:
Adding this support entailed updating tev to the latest (non-backwards-compatible) version of nanogui, which entailed
- porting all tev-specific nanogui features
- making tev agnostic to the rendering API (now supports OpenGL and Metal)
- updating many other dependencies while at it.

This means that there is significant potential for breakage and bugs, so I'm waiting with pushing out a binary release and leaving my fingers crossed that everything will be ironed out by those folks who stay up to date with master. (Thank you very much! :))

In other news: due to the newer nanogui, tev now requires a C++17-capable compiler.


KarelPeeters:
The `nanogui` dependency doesn't compile for me on windows, the function `childCount()` [here](https://github.com/mitsuba-renderer/nanogui/blob/8fe172d32167848b25866672f0ac20f21d6be200/src/widget.cpp#L185) does not exist, that should probably have been `child_count`? Strange that the CI didn't catch this, or maybe I did something wrong.

Tom94:
> The `nanogui` dependency doesn't compile for me on windows, the function `childCount()` [here](https://github.com/mitsuba-renderer/nanogui/blob/8fe172d32167848b25866672f0ac20f21d6be200/src/widget.cpp#L185) does not exist, that should probably have been `child_count`? Strange that the CI didn't catch this, or maybe I did something wrong.

Good catch, thanks! The bodies of assertions are only compiled in debug mode -- yet another example of why C macros are a bad thing.

PR #131, opened 2021-08-01
Expand on inter process communication (IPC)

### LINKED ISSUE #116: IPC docs
First of all thanks for writing this tool, it very quick and easy to use!

I read that tev supports some kind of IPC protocol where it can show an image while it is being rendered. From what I gather you're supposed to send the image data to tev using TCP on the port it prints when it starts up, and the port is also configurable from the commandline options, that all makes sense.

What I can't find anywhere is what exactly I'm supposed to send to tev. Do I just dump the content of an exr file in a single TCP packet? Or is there some kind of custom protocol for sending incremental updates to tev?

Maybe instructions for this could be added to the README? I could be missing something but I can't find any documentation about the IPC feature.

### LINKED ISSUE #120: IPC updateImage performance issue
I was running a render that was pretty fast, about one minute for 1920x1080, data sent to tev in 16x16 blocks of pixels which looks like this:

image

I imagine that this is a typical intented usecase of the IPC. I noticed that tev was using about 25% of the CPU, and some profiling revealed that a lot of the time was spent in setting up and taking down the threadpool here: https://github.com/Tom94/tev/blob/d7aaf6c0f02dfefa0ad7c565bb7da2d33b6a0876/src/Ipc.cpp#L251

Could this threadpool be made persistent? Right now it's created and destroyed for each `updateImage` packet received which is not very efficient.

Tom94:
- Adds an extended version of [Thomas Iser's Python IPC implementation](https://gist.github.com/tomasiser/5e3bacd72df30f7efc3037cb95a039d3)
- Fixes #116 by adding information about IPC to the readme (with a link to [Karel Peeters' Rust implementation](https://crates.io/crates/tev_client))
- Fixes #120 by using a global threadpool in most places rather than local threadpools

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,15 +25,17 @@ _A false-color comparison of two multi-layer OpenEXR images of a beach ball. Ima
 
 ### Graphical User Interface
 
-Images can be opened via a file dialog or simply by dragging them into __tev__. They can be reloaded, closed, or filtered at any time, so don't worry about opening more images than exactly needed.
+Images can be opened via a file dialog or by dragging them into __tev__.
+They can be reloaded, closed, or filtered at any time, so don't worry about opening more images than exactly needed.
 
-Select an image by left-clicking it, and optionally select a reference image to compare the current selection to by right-clicking. For convenience, the current selection can be moved with the Up/Down or the 1-9 keys. For a comprehensive list of keyboard shortcuts simply click the little "?" icon at the top (or press "h").
+Select an image by left-clicking it, and optionally select a reference image to compare the current selection to by right-clicking.
+For convenience, the current selection can be moved with the Up/Down or the 1-9 keys. For a comprehensive list of keyboard shortcuts click the little "?" icon at the top (or press "h").
 
-If the interface seems overwhelming, simply hover any controls to view an explanatory tooltip.
+If the interface seems overwhelming, you can hover any controls to view an explanatory tooltip.
 
 ### Command Line
 
-Simply supply images as positional command-line arguments.
+__tev__ takes images as positional command-line arguments:
 ```sh
 $ tev foo.exr bar.exr
 ```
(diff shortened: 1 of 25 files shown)


src/python/ipc-example.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python3
+
+# This file was developed by Thomáš Iser & Thomas Müller <thomas94@gmx.net>.
```
tomasiser, line 3:
```suggestion
# This file was developed by Tomáš Iser & Thomas Müller <thomas94@gmx.net>.
```

src/python/ipc.py
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env python3
+
+# This file was developed by Thomáš Iser & Thomas Müller <thomas94@gmx.net>.
```
tomasiser, line 3:
```suggestion
# This file was developed by Tomáš Iser & Thomas Müller <thomas94@gmx.net>.
```

README.md
```diff
@@ -43,11 +45,41 @@ By default, all layers and channels are loaded, but individual layers or channel
...
+| Operation | Function
+| :--- | :---------- 
+| `OpenImage` | Opens an image from a specified path on the machine __tev__ is running on.
+| `CreateImage` | Creates a blank image with a specified name, size, and set of channels.
+| `UpdateImage` | Updates the pixels in a rectangular region.
+| `CloseImage` | Closes a specified image.
+| `ReloadImage` | Reloads an image from a specified path on the machine __tev__ is running on.
+
+__tev__'s network protocol is already implemented in the following languages:
+- [Python](src/python/ipc.py) by Thomáš Iser
```
tomasiser, line 70:
```suggestion
- [Python](src/python/ipc.py) by [Tomáš Iser](https://github.com/tomasiser)
```

tomasiser (changes requested):
Thanks a lot for including the Python IPC demo in the project! I am just adding this little review to fix my name spelling 🤭 No worries!

Tom94:
Oh snap, my bad. Too used writing it with h. :p Fixed!

PR #134, opened 2021-08-10
C++20 usage & other improvements

### LINKED ISSUE #123: Keyboard shortcuts: shift does too many things
I'm enjoying tev, thanks for you work 🙏 . I am comparing EXRs by going into 'error' mode and pumping the Exposure using the 'E' key. However decrease exposure is 'Shift-E', but 'Shift' toggles the reference image as well.

Maybe 'Alt-E' could be added as an alternative to 'Shift-E' (or some other workaround)?

### LINKED ISSUE #146: memory leak when the network update function is used
I'm seeing a memory leak in recent versions of tev when sending images from pbrt running another process. (I've been seeing this for a while now, maybe a few months, but only made a repro case now--apologies.)

The attached repro program (extracted from pbrt, so not minimal!) takes a single command-line argument and connects to a running tev server at that address. It then spams it continuously with image updates. I am seeing two things (this is on OSX, FWIW):

* tev's memory use (as reported by top) grows by about 1 MB every second
* interestingly, if I hide tev (command-H), or make another window active and then click on tev to make it active again, memory use jumps up by about 50MB within a few seconds than another 40MB or so again a few seconds later.

So, between those two, it can add up to gigabytes after a while, especially with a long render running.

Tom94:
By using C++20 coroutines, it becomes much easier to effectively utilize tev's threadpool, leading to 2-3x faster parallel loading of large numbers of images.

Also starts using C++20 features in a few minor other places

Recent _other_ improvements to __tev__ were developed on top of this branch. These include
- Visualization of non-trivial data- and display windows #143
- Better multipart EXR support #144 
- Support for the QOI image format #148 (thanks to [Tiago Chaves](https://github.com/laurelkeys))
- And several improvements to __tev__'s startup behavior #145
- Fixes #146 
- Fixes #123 

This PR will stay unmerged until C++20 support became more widespread.


Tom94:
GitHub workflows finally support C++20 on all operating systems, so CI is finally going through! :)

Still, this PR will remain unmerged until C++20-compatible compilers become more widespread. You can track this branch rather than `master` to follow along with __tev__'s development till then.

Tom94:
Merging this now. In case you would like to compile __tev__ and have no way to install a C++20-capable compiler, you can check out the tag `cpp17`. (https://github.com/Tom94/tev/tree/cpp17)

diiigle:
> Disclaimer: Fortunately I'm not a production person and not entirely concerned with backwards compatibility issues. But I still care enough to bring this up here.

All in all, its a bit of a pickle switching to such a 'new' standard. I guess to the fast-moving research community (which this software is probably mostly targeting) this shouldn't pose a problem, but there might be others who can't upgrade that easily. I'm thinking about the rendering people in production, those who might be bound to the  [VFX reference platform](https://vfxplatform.com/). 

Unfortunately it has to this date still no (full) cpp20 compatible compiler. GCC 9.3.1 doesn't seem to support [coroutines](https://gcc.gnu.org/projects/cxx-status.html#:~:text=10%20(requires-,%2Dfcoroutines,-)) with this version yet, but lets the CMAKE_CXX_STANDARD_REQUIRED check pass, because it already supports _some_ features of c++20. (aswf has [docker images](https://github.com/AcademySoftwareFoundation/aswf-docker) for the platform in case you want to add it into your CI workflow)

I understand you don't want to maintain a pure c++17 branch in addition to the cpp20 (master now) branch. But leaving c++17 users without some features but most importantly bugfixes might also be problematic.

Possibly a solution to this problem would be to provide Linux binaries. I know this is a pain in itself which is why you haven't done so in the past, but I did a little research recently, and it seems like with AppImage or Flatpak its getting quite good nowadays. One of the two can also package system libraries (glibc) such that you become truly independent of the executing host system (forgot which one 😺).
I think that preparing the packaging infrastructure would also attract some package maintainers of various Linux distributions, because it'll make their job much easier. CMake (INSTALL) should be of great help with preparing the right folder structure already, then it's only about figuring out the parameters for the packaging tool.

Tom94:
I would love to provide Linux binaries, actually! Let me open an issue that tracks this.

(RE switching to C++20 this early: normally I would completely avoid it for the very same reasons that you quote. Hard agree with all of them. The reason I still pulled the trigger is that coroutines allow very clean code to extract much better parallelism when loading many images, yielding upward of 2x faster startup times. This was too much for me to ignore -- without wanting to turn the existing code into callback hell or spawning hundreds of threads to achieve the same behavior.)

PR #137, opened 2021-09-29
Added link flag for Clang libc++

mcrescas:
Hi :wave: ,

While compiling using `Clang` it is needed to link to `libc++` (the same happens with the `nanogui` dependency). 
The following small addition just check if is the case and set the flags accordingly (code is from @wjakob).

I would like to take this opportunity to thank you for this very useful piece of software!


Tom94:
Hi there; thanks for the PR! :)

Apologies for getting back to you so late. The reason is that I want to take a little time to fool around with Clang/Linux myself before merging -- just to make sure this doesn't break some user's use case (unlikely, but worth making sure).

PR #138, opened 2021-10-08
Fix relative path resolution in background loader

merlinND:
Hi @Tom94,

The worker thread / process used by `BackgroundImagesLoader` does not seem to use the main thread's current working directory.
This leads to relative path resolution when image loading is scheduled by `BackgroundImagesLoader::enqueue()`.

I ran into this issue when using e.g.:

```
tev -n ./image*.exr
```

There's now a bit of code duplication between `tryLoadImage(path, channelSelector)` and `BackgroundImagesLoader::enqueue` (path resolution is done twice). 
Let me know if you'd prefer to avoid it, e.g. by finding a way to set the `cwd` of the worker thread instead (sounds a bit more complicated).

### DIFF
--- a/include/tev/Image.h
+++ b/include/tev/Image.h
@@ -146,7 +146,7 @@ struct ImageAddition {
 
 class BackgroundImagesLoader {
 public:
-    void enqueue(const filesystem::path& path, const std::string& channelSelector, bool shallSelect);
+    void enqueue(filesystem::path path, const std::string& channelSelector, bool shallSelect);
     ImageAddition tryPop() { return mLoadedImages.tryPop(); }
 
 private:
--- a/src/Image.cpp
+++ b/src/Image.cpp
@@ -456,7 +456,16 @@ shared_ptr<Image> tryLoadImage(path path, string channelSelector) {
     return tryLoadImage(path, fileStream, channelSelector);
 }
 
-void BackgroundImagesLoader::enqueue(const path& path, const string& channelSelector, bool shallSelect) {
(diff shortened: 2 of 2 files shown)


merlinND:
Hi Thomas,

This happens to me on macOS, both when trying to load images from a local directory and an SMB directory.

Here's how I got suspicious about the CWD: at the top of `tryLoadImage(path, channelSelector)`, I added:

```c++
    std::cout << "Current directory: " << path::getcwd() << std::endl;
```

And here's what comes out:

```
$ tev -n ./opt*00-00.exr
Current directory: /Users/merlin/Desktop/test-img
16:43:16 SUCCESS  Loaded '/Users/merlin/Desktop/test-img/opt-0000-00.exr' via OpenEXR after 0.046 seconds.
Current directory: /Users/merlin/Desktop/test-img
16:43:16 SUCCESS  Loaded '/Users/merlin/Desktop/test-img/opt-0100-00.exr' via OpenEXR after 0.038 seconds.
Current directory: /path/to/tev.app/Contents/Resources
16:43:16 SUCCESS  Loaded '/Users/merlin/Desktop/test-img/opt-0200-00.exr' via OpenEXR after 0.041 seconds.
Current directory: /path/to/tev.app/Contents/Resources
[etc]
```

(this this after applying the fix for this PR, otherwise image loading fails since the relative paths cannot be resolved when CWD is `/path/to/tev.app/Contents/Resources`).

Maybe the worker mechanism is implemented with a separate process on macOS?

PR #147, opened 2021-11-26
WIP: Add initial support for the QOI image format

laurelkeys:
This PR adds support for the new "Quite OK Image" format ([qoi](https://github.com/phoboslab/qoi)).

---

This PR is currently labeled as ⚠ **work in progress** ⚠ because:
- The QOI format specification will suffer some breaking changes in the following days, after which it should be "locked" (see https://github.com/phoboslab/qoi/issues/28);
- `qoi.h` doesn't seem to have a way for querying the number of color channels in a file when loading it, analogous to stb_image's `int *channels_in_file` output parameter (see https://github.com/phoboslab/qoi/issues/16);
- Also unlike stb_image, `qoi.h` does not have functions that take in some parameter like `stbi_io_callbacks`. Hence, I'm using `path.str().c_str()` directly for loading and saving the files (instead of C++ streams). Please let me know if this is a problem, or if the code can be kept this way.

Other notes:
- I have replaced `filesystem::path&` -> `::filesystem::path&` in the `include/tev/imageio/Image*.h` files, as Visual Studio warns about `filesystem` being ambiguous;
- I have replaced `ostream& iStream` -> `ostream& oStream` in `src/imageio/StbiLdrImageSaver.cpp`, for consistency;
- I have only compiled this on Windows, using Visual Studio 16 2019. It would be great to have it tested on at least one other platform to make sure there won't be problems with `qoi.h` being a C99 file (also, see `src/imageio/QoiImplementation.c`).


Tom94:
Hi there, this is amazing, thank you very much!

I have a small request: __tev__ is currently transitioning to cpp20 with several changes having already been made to the image loader. These are mostly related to supporting async computation via coroutines (`co_await` and `co_return`). (see https://github.com/Tom94/tev/pull/134)

Would it be possible for you to target this PR at the `cpp20` branch rather than the `master` branch to avoid porting effort in the future?

Thanks again in any case!

Tom94:
Regarding your concern about `qoi.h`: I enabled CI for your PRs. This will shed some light on whether things compile on all platforms. I will additionally make sure that there are no runtime errors before merging. :)

laurelkeys:
> Would it be possible for you to target this PR at the `cpp20` branch rather than the `master` branch to avoid porting effort in the future?

Sure! I will fetch the C++20 branch and open a new PR targeting `cpp20` once I make and test the changes. Btw, I think I can use `qoi.h`'s encode/decode functions directly with C++ streams to replace the use of `path.str().c_str()`.

Tom94:
Thanks!

From quickly looking at `qui.h`, the read/write functions copy the entire file into memory anyway, so I agree that doing the copy/reserve on __tev__'s end and subsequently calling decode/encode is better.

That said -- perhaps at some point in the future -- a single-pass over the file would be even better, given that `qui.h` is a fully streaming algorithm.

laurelkeys:
@Tom94 if the next releases of **tev** are going to be based on the `cpp20` branch, I believe this PR can be closed in favor of https://github.com/Tom94/tev/pull/148 (?)

Tom94:
Oh wow, that was quick. Very much appreciated!

PR #152, opened 2021-12-17
#Fix UNIX install prefixes according to CMake recommendation 

diiigle:
This is a small convenience change to CMake installation on Linux. It does not change the current/default behavior, but allows some freedom for changing installation destinations from CACHE variables ("the outside"). 

From https://cmake.org/cmake/help/latest/command/install.html#targets:
> To make packages compliant with distribution filesystem layout policies, if projects must specify a DESTINATION, it is recommended that they use a path that begins with the appropriate GNUInstallDirs variable. This allows package maintainers to control the install destination by setting the appropriate cache variables.

This allows altering of the 'bin' subdirectory by overwriting CMAKE_INSTALL_BINDIR from the outside.

Tested on CentOS with `cmake version 3.20.5 `
Feel free to squash the second commit.


Tom94:
This looks good to me, thank you very much!

I'm currently on vacation without access to most of my machines, so I'm gonna hold off with merging until I can test it on at least one other distro (not foreseeing any problems, just want to make sure).

Cheers!

PR #179, opened 2022-03-15
Change "ctrl" in help text to COMMAND to show correct text in apple.

fjavifabre:
Hi!

I was trying to add some new features in a personal fork when I noticed that some help text is using a RAW "CTRL" string instead of the defined COMMAND variable that uses different strings taking into account the OS, so this is just a small change to fix that.


Tom94:
Hi there, the use of "Ctrl" is intentional, since those particular shortcuts indeed use ctrl, even on macOS. Or did you spot a shortcut that responds to Cmd instead of Ctrl without being labeled as such?

fjavifabre:
Hi. 

No, I just assumed it was supposed to work the other way and it was expected that those "Ctrl" keys changed to Cmd in macOS. I forgot to try all of them to be sure since I had no macOS computer available when I found that.

Sorry!

Issue #181, opened 2022-05-15
Handling duplicate IPC images


anderslanglands:
Say I have a small python script that does something like:

```python
tev.create_image('ref', width=crop_size[0], height=crop_size[1], channel_names=channels)
tev.update_image('ref', image_ref.numpy().reshape(crop_size[0], crop_size[1], 3), channels)
```

first time running this script I will get the 'ref' image created, and it will be updated correctly. Now if I leave tev open and run the script again, I get a second 'ref' image in the image list, which remains black and the first 'ref' image I created gets updated.

There's two options for the behaviour here:
1) When `create_image` is called a second time with the duplicate name, the duplicate call is ignored. This would lead to the current behaviour, except that we wouldn't have empty images created every time we run the script
2) `create_image` could modify the requested name to be unique, and return the uniquified name over the socket. The calling script would then store the returned name and use it in subsequent `update_image` calls to fill in the new image, leaving the first-created one alone.

Option 2 is most similar to behaviour in common render catalogs (e.g. it, katana's monitor etc), but given that there isn't currently a mechanism for returning data to the client over the socket, it's probably simpler just to do 1) for now. I'm happy to do this and submit a PR if you're open to it.

PR #182, opened 2022-05-15
Don't create duplicate images over IPC

anderslanglands:
As discussed here: https://github.com/Tom94/tev/issues/181 this PR adds a check in IPC that skips creating a new image if an image with that name already exists.


Tom94:
Thank you very much! I agree very much with the points raised in the corresponding issue. This is a much better approach than the previous "duplicate image" policy.

PR #191, opened 2022-10-08
Replace non-ASCII

chaosink:
MSVC reports an error "newline in constant" for `"…"s`.

The non-ASCII hyphens cannot show correctly in the window title.


Tom94:
Hi, the non-ASCII characters were intentional and show correctly on all systems I've tested __tev__ on. Rather than changing them, I'd like to find out more about your system and the failure conditions. Could you share the exact error MSVC gives you + any info about your system that you can think of might break non-ASCII characters?

PR #192, opened 2022-10-16
Specify UTF-8 charset for MSVC

chaosink:
On Windows systems with non-English locales, the UTF-8 chars may cause "Error C2001: Newline in constant".

2 solutions:
    Save the file with UTF-8 with BOM (byte-order mark).
    Specify UTF-8 charset for compilation. Used by this commit.

See discussions here:
    https://github.com/libusb/libusb/issues/207#issuecomment-288664124

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,6 +113,10 @@ if (MSVC)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4")
     endif()
 
+    # tev uses UTF-8 chars. Specify UTF-8 charset explicitly.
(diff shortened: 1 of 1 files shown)


Tom94:
This is great, thank you! Before I hit merge, could you move the code to the existing `if (MSVC)` block that already adds a few other flags? I'm thinking somewhere around line ~115.

Tom94:
I'd prefer `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /utf-8")` for consistency with the other flags that are set in the block around 115.

I'm all for migrating to `target_compile_options()`, but imo this is better done as an all-or-nothing thing, also using `target_include_directories` and fixing up dependencies etc.

Tom94:
Thanks again!

PR #193, opened 2022-10-16
Use key Tab to show pixel raw bytes

chaosink:
Shift+Ctrl should be held to show pixel raw bytes.
But this keybinding conflicts with others, like reference image selection.
A usage case prevented by this is that: keep showing pixel raw bytes, and switch between images by number keys for comparison.

This commit changes the keybinding to Tab.
@Tom94 Maybe you want another key.
Thanks!

### DIFF
--- a/src/ImageCanvas.cpp
+++ b/src/ImageCanvas.cpp
@@ -137,9 +137,7 @@ void ImageCanvas::drawPixelValuesAsText(NVGcontext* ctx) {
         nvgTextAlign(ctx, NVG_ALIGN_CENTER | NVG_ALIGN_MIDDLE);
 
         auto* glfwWindow = screen()->glfw_window();
-        bool shiftAndControlHeld =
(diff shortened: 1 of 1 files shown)


Tom94:
This is a good idea. I've been bothered by the overlap in hold-keybindings for a while... although what kept me away from Tab so far is how easy it'd make it to press Ctrl+Tab or Alt+Tab accidentally.

However, now that I think about this a bit more deeply: the spacebar is currently unused in __tev__, and I can't think of typical OS-wide shortcuts that it might cause conflicts with. What's your opinion on using space instead of tab?

(That aside, for the PR, could you also change the listed keybinding in `HelpWindow.cpp`?)

Tom94:
Oh wow, I totally forgot about playback -- thanks for pointing it out. You're right, I need to add those to the help window. And while at that make it scrollable... IIRC it's already pushing the default size of the window.

I do like 'B'. While it conflicts with Ctrl+B (which hides/shows the sidebar), that shortcut is only triggered when Ctrl is already held at the point when B is pressed. Which I suppose happens rarely enough. Please feel free to make the change.

Issue #164, opened 2022-01-06
tev should release Linux binaries


Tom94:
Linux binaries would be good to have, especially now that __tev__ requires a C++20-capable compiler, which many people do not have access to in their work environments. (see also https://github.com/Tom94/tev/pull/134)

Things on the TODO list:
- Look into AppImage and Flatpak
- Look into getting binaries into package repositories

I currently do not have cycles, so it will be some time until there is an update from my end, but if anyone is familiar with releasing Linux binaries all help is much appreciated!

Calinou:
I might look into creating a Flatpak of tev, but I don't know if the latest FreeDesktop SDK can compile C++20 code.

Tom94:
We've got x64 Linux binaries now; thanks @diiigle! Low priority for future work is to also make available ARM binaries.

PR #194, opened 2022-11-05
Package linux binaries as AppImage

diiigle:
I had some time to get into AppImage and used `tev` as my demo application.

The solution calling CPack is inspired by [this gist](https://gist.github.com/bruxisma/019106e20b297174abf30a8f74eff9b3)
It introduces an "External" CPack Generator script that (downloads and) calls `linuxdeploy` on an installed directory (by CMake) structure that adheres to the *AppDir* specification.

An attempt to solve #164

I was not able to test it on a Linux machine with a display attached yet, but it builds in the CI (tested) and starts (until it can't find a display). So I don't know if the icon shows up on the file and if the `.desktop` integration works.


Tom94:
Hi, this is great, thank you very much!

__tev__ also runs on 32-bit and ARM linux, which will break the hardcoded x86_64 download. Since linuxdeploy does not release ARM binaries, would it be possible to make linuxdeploy a submodule and to build it for per cmake before using it in cpack?

diiigle:
Good catch. 

ARM support for the linuxdeploy CI seems to have stalled https://github.com/linuxdeploy/linuxdeploy/issues/175. But the main  AppImage tool (`AppImageKit`) seems to have a [running arm(64) CI by the same maintainer](https://github.com/AppImage/AppImageKit/blame/continuous/.github/workflows/build.yml). 
I'm torn if cooking a special solution locally or fixing it upstream would make more sense.
I'll have a peek into the build implications of linuxdeploy first. 

Does `tev` have (self-hosted) runners for arm and 32bit?

Tom94:
There are no runners, but I wouldn't mind building the appimages per-release manually on a VM (at least for ARM). :)

If it's not as simple as `add_subdirectory(dependencies/linuxdeploy)` (and maybe a couple lines of setup), then don't worry about it. That aside, I think it'd be nicer to have linuxdeploy part of the build chain rather than a downloaded binary, even if it only works on x86 for now.

Tom94:
Merging this for now -- arm linux can be future work. Thanks again!

diiigle:
I was/am actually looking into it. Will probably find time tomorrow again.

Building linuxdeploy from scratch is a bit of a rabbithole, as you also need to build the "appimage plugin". But I should manage.
In any case, 32bit bit support is also missing the way you have merged it now.

Tom94:
That's great to hear. Even with arm & 32-bit missing, it's still great to have what you've already done! Another PR is of course much appreciated, but don't feel pressured. :)

Issue #175, opened 2022-02-07
In network mode, highlight last received packet/block


renderdude:
Most viewers when used with a progressive renderer, have some way of highlighting the current pixel/tile/block that is being rendered (Pixar's IT uses a blue box, blender uses crosses at the corners).

When using tev (very nice, thank you) with PBRT, as the render takes longer per tile (more samples, ....) it becomes difficult to impossible to identify where in the image pbrt is rendering.

Is it possible to add an indicator for the last received set of pixels. that is erased when the new pixels arrive and a new indicator is displayed?

Tom94:
Hi, sorry for the delay in getting to this.

A good way to achieve this would be to add network packets for drawing vector graphics on top of a given image. This way, the rendering software itself can choose how it would like to highlight the in-progress rendering, rather than __tev__ interfering with the image in a potentially undesired way.

PR #200, opened 2022-11-29
Vector graphics over IPC

Tom94:
Adds support for sending vector graphics commands over IPC to annotate images.

For example, this can be used to highlight the latest rendered part of an image (supposing that __tev__ is used as the frontend for a renderer). You can view a sample of how this can be accomplished by running `src/python/ipc-example.py`. Another use case would be marking NaNs, infinities, or otherwise unexpected values, depending on context.

Implements #175 

@mmp let me know if this is interesting for you and you have early feedback. :)

Here's sample Python code to draw a rectangle:
```python
import tev
with tev.Ipc() as tev_ipc:
    # Assumes that an image named "image" has been opened or created beforehand
    tev_ipc.update_vector_graphics("image", [
        tev.vg_begin_path(),
        tev.vg_rect(x, y, width, height),
        #Alternatively: draw rectangle manually
        # tev.vg_move_to(x, y),
        # tev.vg_line_to(x, y + height),
        # tev.vg_line_to(x + width, y + height),
        # tev.vg_line_to(x + width, y),
        # tev.vg_close_path(),
        tev.vg_stroke(),
    ])
```

Here's sample C++ code to draw a rectangle:
```cpp
#include <tev/Ipc.h>
#include <tev/VectorGraphics.h>

using namespace tev;

Ipc ipc; // Will connect to running instance of tev if one exists
if (!ipc.isPrimaryInstance()) {
    IpcPacket packet;
    
    // Assumes that an image named "image" has been opened or created beforehand
    packet.setVectorGraphics("image", false, false, {
        VgCommand::begin_path(),
        VgCommand::rect(x, y, width, height),
        //Alternatively: draw rectangle manually
        // VgCommand::move_to(x, y),
        // VgCommand::line_to(x, y + height),
        // VgCommand::line_to(x + width, y + height),
        // VgCommand::line_to(x + width, y),
        // VgCommand::close_path(),
        VgCommand::stroke(),
    });
    ipc.sendToPrimaryInstance(packet);
}
```


mmp:
Nice! That's an impressive amount of functionality available (Beziers even!). No immediate feedback but I'll think about uses of it (and try some experiments.) For the use case of nan/inf, I'd argue that tev is better placed to report those, since it has more flexibility in how it displays information and since it sees them just the same as the renderer. So I think the bigger value is from things that only the renderer knows; beyond "I just finished these pixels", I'll think more about other things like that that could be interesting...

PR #201, opened 2022-12-05
Linux AppImage packaging v2

diiigle:
I finally got around to finish my work on the AppImage packaging (following up on #194, #164).
This at least adds support for `x86` builds, for `arm` there are still some downstream problems (linuxdeploy/linuxdeploy#189) in `linuxdeploy` itself.

The three commits:
 1.  Cleaning up the CMake a bit by moving the CPack script to a separate file. Then using the host architecture instead of the previously hardcoded `x86_64`.
 2. & 3. are adding `linuxdeploy` (`continuous` branch) as a submodule and building it as a dependency of tev. It takes quite some additional time to build that, and some sub-sub dependencies are still being downloaded as binaries from github. 

The first commit is worth merging IMHO, the other two are up to your consideration. It *should* work on its own as well - the (potentially) downloaded binary has the same name as the one build from sources.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -249,47 +249,8 @@ else()
         install(FILES resources/icon-${RES}.png DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/${RES}x${RES}/apps" RENAME tev.png)
     endforeach()
 
-    file(GENERATE
-        OUTPUT "${PROJECT_BINARY_DIR}/appimage-generate.cmake"
-        CONTENT [[
-include(CMakePrintHelpers)
-cmake_print_variables(CPACK_TEMPORARY_DIRECTORY)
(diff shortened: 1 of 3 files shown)


diiigle:
Reported the failing build. With some modifications it builds locally for me.
https://github.com/linuxdeploy/linuxdeploy/issues/212#issuecomment-1337263492

Tom94:
Hi again; thanks so much! I'd be happy to merge the first commit (and, I guess, the README update) as-is if you'd like to make a separate PR with those.

If linuxdeploy adds minutes to the build process _and_ runs into unforeseen trouble (as in CI), it's probably not worth the effort to bundle. Thanks again looking into it.

diiigle:
Okay, I dropped the respective commits and the CI is building now.

Tom94:
Thanks again!

PR #208, opened 2023-04-20
Add textbox to edit image caption in the image list

Speierers:
# Overview

This PR adds the ability to edit the caption of the selected image in the image list via a new textbox added to the GUI.

image

A keybinding for the `Enter` key was added to trigger the focus on the caption text box for users to quickly annotate different images in the list. We also ensure that captions are preserved when reloading an image.

## Example

The following screencast showcases this new feature:

https://user-images.githubusercontent.com/6662561/233287487-238a572e-e0dd-4f35-8934-c159fc3b5899.mov

### DIFF
--- a/src/ImageViewer.cpp
+++ b/src/ImageViewer.cpp
@@ -1060,6 +1060,10 @@ void ImageViewer::insertImage(shared_ptr<Image> image, size_t index, bool shallS
         }
     });
 
+    button->setCaptionChangeCallback([this]() {
+        mRequiresFilterUpdate = true;
+    });
+
     mImageButtonContainer->add_child((int)index, button);
     mImages.insert(begin(mImages) + index, image);
 
--- a/src/ImageViewer.cpp
+++ b/src/ImageViewer.cpp
@@ -450,20 +450,6 @@ ImageViewer::ImageViewer(const shared_ptr<BackgroundImagesLoader>& imagesLoader,
 bool ImageViewer::mouse_button_event(const nanogui::Vector2i &p, int button, bool down, int modifiers) {
     redraw();
 
-    // Hide caption textbox when the user performed mousedown on any other component
-    if (down) {
-        nanogui::Vector2i relMousePos = (absolute_position() + p) - mImageButtonContainer->absolute_position();
-        auto& buttons = mImageButtonContainer->children();
-        for (size_t i = 0; i < buttons.size(); ++i) {
-            ImageButton* imgButton = dynamic_cast<ImageButton*>(buttons[i]);
-            if (!imgButton->contains(relMousePos)) {
-                imgButton->hideTextBox();
-            } else if (imgButton->textBoxVisible()) {
-                return true;
-            }
-        }
-    }
-
     // Check if the user performed mousedown on an imagebutton so we can mark it as being dragged.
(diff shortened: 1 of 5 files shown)


wjakob:
What about hiding this text box by default or realizing the feature using a hotkey-triggered modal dialog? (tev already has so many buttons and widgets on the left hand panel, and this adds further elements)

Tom94:
Thanks a bunch for the contribution, Sebastian!

I concur with Wenzel in that the additional UI element feels unnecessary. Could you make the image button turn into a textbox upon pressing the hotkey? This would signify to the user what they're actually editing here -- like this:
https://user-images.githubusercontent.com/4923655/233303095-2020f7f8-9d50-4319-8a4e-f8db18f4c4d7.mov

Also, could you make the shortcut be F2 on Windows and Linux (which is the common "rename" shortcut on those systems) and list the shortcut in `HelpWindow.cpp`?

Speierers:
I agree, let's keep the UI clean.

I modified my patch so that it create a `TextBox` in the image list when pressing F2 to edit the image caption. To cancel, the user can simply click anywhere else on the application.

https://user-images.githubusercontent.com/6662561/233362098-a000c765-c084-4778-a5a0-198ae58e0cae.mov

src/ImageViewer.cpp
```diff
@@ -1475,6 +1501,49 @@ void ImageViewer::setGamma(float value) {
...
+    mCurrentCaptionTextBox->set_editable(true);
+    mCurrentCaptionTextBox->set_alignment(TextBox::Alignment::Left);
+    mCurrentCaptionTextBox->set_placeholder("Enter caption");
+    mCurrentCaptionTextBox->set_callback([this](const string& name) {
+        if (!mCurrentImage) {
+            return true;
+        }
+
+        int id = imageId(mCurrentImage);
+        ImageButton* ib = dynamic_cast<ImageButton*>(mImageButtonContainer->children()[id+1]);
```
Tom94, line 1522: This is brittle and prone to errors if tev keeps loading and/or closing images while a caption is being edited. I think it's hopeless to catch all those cases and make them close the textbox -- instead, caption editing should be robust to changes in the images (even to the image currently being edited getting closed).

If I were to tackle this, I'd make the text box a child of the image button in question, moving much of the logic into `ImageButton.cpp` and `ImageButton.h`. That way, the text box automatically disappears if the image does, and the caption can be edited without having to hold a reference to the image itself. (This would also make my above comment obsolete.)

Tom94 (changes requested):

Tom94:
That looks very sleek. Thanks!

Aside from UX, I've some additional concerns regarding the underlying implementation. Once those are solved this should be good to merge.

Tom94 (approved):

Tom94:
Thanks! I changed / added a few additional things and will merge once CI is through.
- styling of the textbox
- default value is now the previous name (which is fully selected upon rename start)
- simplified `ImageViewer`-side code to cancel the edit
- added filter/highlight updates upon rename completion
- complete the rename, even if canceled by clicking somewhere else

https://user-images.githubusercontent.com/4923655/233633152-125f4b8e-7735-48f1-8b89-f780cd5e594f.mov

PR #214, opened 2023-12-28
Use scientific notation in MultiGraph

iRath96:
Hi,
first of all thanks for this great image viewer!

Currently the MultiGraph uses `%.3f` format to show minimum/mean/maximum, which can be problematic when the image values are small (most notably when comparing error between similar images). This PR changes the format to `%.2e`, which retains the same precision as before, but includes the exponent.

Thanks!

### DIFF
--- a/src/MultiGraph.cpp
+++ b/src/MultiGraph.cpp
@@ -17,7 +17,7 @@ namespace tev {
 
 string formatNumber(float v) {
     bool needsScientificNotation = v != 0 && (abs(v) < 0.01f || abs(v) >= 1000);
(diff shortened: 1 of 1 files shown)


Tom94:
Hi Alex, likewise thanks for the contribution!

I think numbers in the range 0.01 <= abs(x) < 1000 are better recognized (and compared with each other) using the existing 3 decimals. It's also what current tev users are used to.

To maintain compatibility with those, could you change it so that formatting only switches over to scientific outside of that range? I'd also appreciate a special case for x == 0 that maintains the existing format. Thanks!

Issue #229, opened 2024-05-23
Ctrl+number cannot select channel groups on Linux/Windows


MKimiSH:
I believe the cause is the `!(modifiers & SYSTEM_COMMAND_MOD)` in the following line: https://github.com/Tom94/tev/blob/13352738362d65d1b89dea4b81d8756bb5cc1368/src/ImageViewer.cpp#L686. On Windows and Linux, this is the same as Ctrl, so it blocked the Ctrl+number key combination. This seems related with introducing Ctrl/Cmd+0 and Ctrl/Cmd+9 for setting image transforms. 

I think the new Ctrl+0/9 behavior is more useful than selecting the 9th or 10th channel group. Meanwhile, we can perhaps keep Ctrl+[1-8]? It introduces a bit of inconsistency though.

Tom94:
Oops, good catch! Thanks & sorry for the accidental breakage. Just pushed a fix that makes Ctrl+1-8 work (I agree 9 and 0 aren't necessary here).

PR #230, opened 2024-06-02
Add a link in README.md pointing to the keyboard shortcut visualization

MKimiSH:
The motivation behind is that the set of key bindings of tev has really grown and sometimes it's hard to keep track for users and developers (#123, #193, #229). Therefore, I translated the list of hot keys from the help popup, as well as from `ImageViewer.cpp`, to a visualizer I encountered, KeyCombiner, and mentioned it in the README (it's [here](https://keycombiner.com/collecting/collections/shared/f050cc02-f23a-425c-b032-b4c3659c7ef4) for quicker reviewing). It allows viewing the hot keys as a list and see the bindings associated with each modifier (or modifier combination). 

Please feel free to check whether the key bindings are correct and complete. 

PS: 
While doing the translation, I noticed that some of the key bindings are not actually in the help window. For example, the `pageUp` and `pageDown` keys are mapped as `up` or `down` but were not mentioned. We can perhaps update the help window after this PR has been merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Images can be opened via a file dialog or by dragging them into __tev__.
 They can be reloaded, closed, or filtered at any time, so don't worry about opening more images than exactly needed.
 
 Select an image by left-clicking it, and optionally select a reference image to compare the current selection to by right-clicking.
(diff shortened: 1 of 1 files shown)


Tom94:
Thank you, this looks great! Apologies for the delay in getting to the PR.

PR #258, opened 2025-02-25
Switch license to GPLv3

Tom94:
As I'm adding features to **tev**, more and more dependencies become more ergonomic to use if tev is under the GPLv3 license. Over the years, I've also gotten more and more appreciative of the values that free-software licenses represent, so this is a good opportunity to make the switch. tev's version will get bumped to 2.0 to reflect this major change in licensing.

What does this mean in practice?

## For users

__Nothing changes.__ You can keep using **tev** exactly like you did so far, regardless of whether you install a binary or compile from source.

## For developers
- If you directly use **tev** source in your tools, or you statically link against it, __you must either release your tool's source code under the GPLv3 as well, or you must use **tev** code from version 1.30 or earlier, which remains BSD 3-clause licensed.__
- If you dynamically link against **tev** or if you communicate with **tev** over the IPC protocol, __nothing changes__.


MKimiSH:
Thanks for the explanation, that's very clear!

PR #282, opened 2025-04-30
Add JPEG XL Support

zalo:
JPEG XL was the format of the future!!   Now it is the format of the present 😎 

JPEG-XL supports alpha, animations, 32-bit precision, 4100 channels, and images over 1 billion pixels wide!   
All while remaining on the pareto frontier for encode/decode speed to compression ratio.
Learn more about this on Cloudinary's Blog: https://cloudinary.com/blog/jpeg-xl-and-the-pareto-front

I'm considering using it as a total replacement for EXR in my work.   There is now a [great Python library supporting it](https://pypi.org/project/imagecodecs/).   The next step is for a great viewer to support it... that is where this PR comes in.

This PR is a work in progress for properly pairing JPEG-XL's features with tev's to properly inspect and appreciate how far image compression has come since the humble jpeg.

JPEG-XL Platform Chart:
- [x]  Platform: Windows 
- [x] Platform: Mac ARM
- [x]  Platform: Mac Intel
- [x] Platform: Linux 

Features:
- [x] Image Loading/Saving
- [x] Alpha Support
- [x] Color Space Support 75%
- [x] >4 Channel Support
- [x] Animation Support

The images here are useful for testing:
https://github.com/libjxl/conformance
https://jpegxl.info/images/anim-icos.jxl
https://jpegxl.info/images/dice.jxl
https://jpegxl.info/images/Webkit-logo-P3.jxl

@Tom94


Tom94:
Hi there; thanks for the PR! There's quite a bit of work and cleanup still necessary, but judging from your TODO list, it looks like you're on top of most important points.

I'd maybe add that most code comments are redundant and there a bit of spaghetti code duplication here and there. Possibly remnants of the initial vibe coding approach, but nothing that can't be cleaned up later.

A bit more concerning are the extra packages that need a system-wide install. I'd prefer if tev remained buildable without any external `find_package` requirements, i.e. the local lcms should get used and new dependencies like brotli should become a submodule and get built alongside tev.

Eager to follow along!

zalo:
Thanks!

The external packages were all remnants of trying out the "recommended" way of getting `libjxl` into a project 😅 

Since switching to the submodule paradigm, `brotli`, `lcms2` etc. are now all being loaded from the `third_party` subfolder within the `libjxl` submodule.
(For `lcms2`, this is important to note, since it means `libheif` will be using `libjxl`'s version of `lcms2` when `libjxl` is present, since CMake throws a fit when they try to co-exist. 💀  It seems to work on all platforms 😄 ).


With animations, alpha, and channels squared away, it seems like all that remains is:
- Properly interpreting color spaces
- Tricking `libjxl`'s `lcms2` to not use SSE without putting that ugly define at the top of the main CMakeLists.txt...

Also, I'm wondering if the excessive number of warnings generated from compiling `libjxl` are what is causing the MacOS ARM runner to time/glitch out... should I disable warnings for dependencies?

scripts/create-dmg.sh
```diff
@@ -14,7 +14,7 @@ cmake \
     -DCMAKE_BUILD_TYPE=Release \
     "$@" \
     ../.. || exit 1
-make -j || exit 1
```
Tom94, line 17: If this ends up fixing the problem, I think the best solution would be to set the parallelism to the number of cores the machine has. `make -j` can spawn an excessive number of threads -- more than there are cores available -- sometimes leading to OOM.

The way I'd do this is to change the build script to entirely ditch makefiles and use ninja instead :p. I.e. add -GNinja to the configure command and replace the `make` call with `cmake --build . --config Release -j`

zalo: I was seeing `clang: error: unable to execute command: posix_spawn failed: Resource temporarily unavailable` pop up sporadically, so I figured I'd try [this StackOverflow answer](https://stackoverflow.com/a/46255347).

My secondary guess is that Github's ARM Runners are a hot commodity, and each one is stressed out sharing a bunch of jobs 😅 

It looks like it did work!   And it built like 4 times _more quickly_ 🎉: https://github.com/zalo/tev/actions/runs/14750446001/job/41406464717

zalo:
> [I]f libjxl's dependencies are self contained, I'd recommend looping it in via ExternalProject_Add rather than via add_subdirectory. That'll isolate its build from the rest of tev, meaning warnings won't be an issue, you can disable SEE just for it, and there won't be any lcms2 conflicts.

Cool; I'll consider this technique; I might be able to filch some of the template code from here:
https://github.com/uyjulian/ifjxl/blob/main/CMakeLists.txt#L30-L84

> Regarding SSE, though: disabling it unconditionally (even if just for libjxl) ought to have negative consequences for decode speed, right? Probably best to only disable it when targeting arm.

That define is just for the `fast_math` plugin in `lcms2` ([which is currently always set regardless of ARM-ness on `master` 😅](https://github.com/Tom94/tev/blob/master/dependencies/CMakeLists.txt#L156) ).  I don't think it's hurting anything by being there, but its positioning there is quite ugly.

Tom94:
> > [I]f libjxl's dependencies are self contained, I'd recommend looping it in via ExternalProject_Add rather than via add_subdirectory. That'll isolate its build from the rest of tev, meaning warnings won't be an issue, you can disable SEE just for it, and there won't be any lcms2 conflicts.
> 
> Cool; I'll consider this technique; I might be able to filch some of the template code from here: https://github.com/uyjulian/ifjxl/blob/main/CMakeLists.txt#L30-L84

Sounds good; thanks! But maybe check out the way tev handles libaom on latest `master` first -- the one you linked looks like it has some stuff hardcoded that might break on Windows or multi-configuration builds (e.g. the fixed .a paths).

scripts/create-dmg.sh
zalo: 🐌 Race started between:
```machine_type=$(uname -m)
if [[ "$machine_type" == "arm"* ]]; then
  make || exit 1
else
  make -j || exit 1
fi
```
https://github.com/zalo/tev/actions/runs/14750886621

And
```
cmake \
    -GNinja \
    -DCMAKE_OSX_DEPLOYMENT_TARGET=10.15 \
    -DCMAKE_BUILD_TYPE=Release \
    "$@" \
    ../.. || exit 1
cmake --build . --config Release -j || exit 1
```
https://github.com/zalo/tev/actions/runs/14751011995

Winner gets it 😄

zalo: Conditional `make -j`:
image

Ninja:
image

Seems like it _does_ make ARM a lot faster...

Tom94: There's probably also a lot of measurement noise involved, but imho the ninja solution is the cleanest either way. I've yet to run into a cmake-based project where ninja isn't just an all-around better build system than make.

zalo: Alrighty; I'll mark this conversation as resolved then 👍

zalo:
> [I]f libjxl's dependencies are self contained, I'd recommend looping it in via ExternalProject_Add rather than via add_subdirectory.

Hmm, I'm noticing something sketchy with the new `ExternalProject` aom build files...
https://github.com/zalo/tev/actions/runs/14751832507/job/41410882667#step:3:1516

Since the build speed increased and the warning spam largely went away with the switch to Ninja (and I'm not sure I have the skills to make the `ExternalProject` work with `libjxl`), I think I'll hold off on that one and just focus on color profile handling tomorrow...

I notice the colors look correct when I can set the Gamma to 1.0, so hopefully there's some simple transformation I can do for that...

Tom94:
> Hmm, I'm noticing something sketchy with the new `ExternalProject` aom build files... https://github.com/zalo/tev/actions/runs/14751832507/job/41410882667#step:3:1516

Already pushed a fix 1 or 2 hours ago -- macOS version wasn't getting passed along. I'm also still new to it, so might have to follow up with a few more hotfixes.

> Since the build speed increased and the warning spam largely went away with the switch to Ninja (and I'm not sure I have the CMake skills to make the `ExternalProject` path work with `libjxl`), I think I'll hold off on that one and just focus on color profile handling tomorrow...

No hurry! Thanks a bunch for all the work you've already put into this. And happy to give the ExternalProject jxl combo a try myself to get more familiar with it.

Tom94:
> > Regarding SSE, though: disabling it unconditionally (even if just for libjxl) ought to have negative consequences for decode speed, right? Probably best to only disable it when targeting arm.
> 
> That define is just for the `fast_math` plugin in `lcms2` ([which is currently always set regardless of ARM-ness on `master` 😅](https://github.com/Tom94/tev/blob/master/dependencies/CMakeLists.txt#L156) ). I don't think it's hurting anything by being there, but its positioning there is quite ugly.

It being disabled in just tev is fine, because the SSE-relevant bits of the fast float plugin aren't used anywhere (tev only uses it for fp32<->fp32 conversions). But jxl might be different. I didn't check, but I could imagine the fp32<->8bit code path of lcms' fast_float that's actually accelerated by SSE to be in use there.

Tom94:
Okay, I had a go at ExternalProject and, while it worked, it turned out way too convoluted. Much worse than the current situation. 😅

I took the liberty to clean up lcms2 and a few minor CMake things while at it. Didn't touch any other part of the PR. Settled on letting lcms2 build twice -- once for jxl and once for tev w/ fast_float enabled. It's fairly fast to compile, so little downside to the extra isolation.

zalo:
Awesome; thank you!  As a CMake newbie, I greatly appreciate your advice and guidance 😄

I’m hoping I can finish up the rest of this PR today 😎

Todo for myself:
- Add File Selector Support
- Reduce JPEGXL Signature Check to 16 bytes
- Consider only decoding animation frames until the host device is about to run out of memory (or another technique to limit a massive file’s memory consumption…)
- Make the color space work without messing with the original pixel values (for the histogram etc.)
- Handle Single Channel (Grayscale) more intelligently
- Consider modeling remaining functionality after imagecodecs:
https://github.com/cgohlke/imagecodecs/blob/8776afc59c94b2531aaecb5dd969308388c9e2d0/imagecodecs/_jpegxl.pyx#L577
- Use basic info’s bitspersample and exponentbitspersample to disambiguate the source bit depth; consider adjusting as needed for tev…
- Figure out what isPlanar from imagecodec’s implementation means 😅

Tom94:
> Awesome; thank you! As a CMake newbie, I greatly appreciate your advice and guidance 😄

🙇‍♂️ FWIW, I also feel very much like a CMake noob. Plus, much of tev's CMake code is old and, frankly, bad. Definitely not good guidance material. 😅 (Notably dependencies could be handled much better by the various `target_*` functions instead of pushing path and compiler variables into parent scope.)

> * Make the color space work without messing with the original pixel values (for the histogram etc.)

Internally, tev represents all images using rec709 primaries, linear colors (gamma == 1), with premultiplied alpha. So it's fine to directly mess with the pixel values.

Setting the `.hasPremultipliedAlpha = false` or `.toRec709` in the load function merely causes a postprocess downstream. So... if you can use something like lcms2 (analogously to how `HeifImageLoader` works) to directly convert the loaded image via icc profile to rec709, linear, premultiplied, then that's as ideal as it gets.

There's also bit depth handling in `HeifImageLoader` that, based on a cursory read of jxl's docs, you should be able to carbon copy.

I'll follow up with a few more comments in the code, 1 sec.

src/imageio/JxlImageLoader.cpp
```diff
@@ -0,0 +1,233 @@
...
+            result.emplace_back();
+            ImageData& data = result.back();
+
+            // Set dimensions
+            Vector2i size{(int)info.xsize, (int)info.ysize};
+            data.dataWindow = {
+                {0,        0       },
+                {size.x(), size.y()}
+            };
+            data.displayWindow = data.dataWindow;
```
Tom94, line 174: You can safely remove these as well. They'll be set to the image size automatically if unset.

At the moment, they're only really useful for EXR images, where the display region of an image can be configured separately from the region that actually has data.

src/imageio/JxlImageLoader.cpp
```diff
@@ -0,0 +1,233 @@
...
+                throw invalid_argument{fmt::format("No channels match '{}'.", channelSelector)};
+            }
+
+            // Copy pixels to channels
+            co_await ThreadPool::global().parallelForAsync<int>(
+                0,
+                size.y(),
+                [&](int y) {
+                    for (int x = 0; x < size.x(); ++x) {
+                        size_t pixelIdx = ((size_t)y * size.x() + x) * numChannels; // RGBA format from JXL decoder
```
Tom94, line 205: This index should be, roughly,
```
size_t align = std::max(format.align, (size_t)1); // Handle case where format.align == 0
size_t bytesPerRow = size.x() * sizeof(float) * numChannels;
bytesPerRow = (bytesPerRow + align - 1) / align * align;
size_t pixelIdx = y * bytesPerRow / sizeof(float) + x * numChannels;
```

From `JxlPixelFormat`'s definition:
```
  /** Align scanlines to a multiple of align bytes, or 0 to require no
   * alignment at all (which has the same effect as value 1)
   */
  size_t align;
```

Tom94: Following up: I just realized you're the one in control of `format.align`, making the above comment moot. 😅

zalo: It might still be useful for the other bit depths or doing some kind of bulk memcpy 😄

Speaking of other bit depths: Is there any scenario where tev would want the images in uint8 or uint16 if the underlying image reported itself that way?

zalo:
Hmm, I added preliminary color handling... but it's a bit messy for the moment...

The majority of images just needed `toLinear()` called on them, so that is nice...  The trouble comes from non-sRGB images.

Here is the Rec 2100 test image I'm using: https://jpegxl.info/resources/HDRSample_Images/sunrise-PQ.jxl

At first I tried using `toRec709()`, but I couldn't get it to apply properly despite supplying coefficients both from the image and coefficients that I know are correct from the UltraHDR's Rec 2100 implementation.   Likewise, no combination of `convertToRec709()` and `toLinear()` would make it look like Apple's preview without adjusting Gamma and Exposure.

Next, I tried using libjxl's "Default CMS" (probably lcms), to request the decoder output a pre-linearized color profile.   This works for most cases, but crashes on the animated icosahedron. 💀    Additionally, the HDR image appears to be scaled differently from Apple's Preview, but it looks right with a 16x multiplier in there (possibly from 8-bit to 12-bit?)

So my temporary solution is to only use the CMS to pre-linearize only non-sRGB images, and to add the 16x multiplier when it detects data in uint16 and a non-standard color encoding.  This seems to work for all of my test images, but I recognize it's pretty brittle...   Ideally I could figure out how to use toRec709 properly AND why libjxl's cms conversion...

I'll comb through `HeifLoader` to solve this soon...

Tom94:
Thanks a bunch for all your work! I had a look at [jxl/color_conversion.h](https://github.com/libjxl/libjxl/blob/main/lib/include/jxl/color_encoding.h) and, yeah... 100% agree that it's best to offload conversion to their and/or a 3rd party cms.

Possibly, we can work around the crash by using lcms2 ourselves from within tev. Might even give us a speedup since jxl doesn't seem to be using its fast_float plugin. (They actually can't, because they're BSD-3 licensed. fast_float is GPLv3 unless you pay for a commercial license, so it's incompatible with them.)

For the 16x multiplier: I agree it's likely 12-bit data that's been put in a 16 bit buffer and hasn't been correctly multiplied by `2**16/2**12` (the multiplier doesn't come from using more than 8 bits but rather fewer than 16).
```
  /** This is the default setting, where the encoder expects the input pixels
   * to use the full range of the pixel format data type (e.g. for UINT16, the
   * input range is 0 .. 65535 and the value 65535 is mapped to 1.0 when
   * converting to float), and the decoder uses the full range to output
   * pixels. If the bit depth in the basic info is different from this, the
   * encoder expects the values to be rescaled accordingly (e.g. multiplied by
   * 65535/4095 for a 12-bit image using UINT16 input data type). */
  JXL_BIT_DEPTH_FROM_PIXEL_FORMAT = 0,
```

Tom94:
Oh, and I almost forgot: if you haven't already found these, here is big batch of official test images along with references: https://github.com/libjxl/conformance

zalo:
I spent some time tracking the crash loading [anim-icos.jxl](https://jpegxl.info/images/anim-icos.jxl) down to here: https://github.com/libjxl/libjxl/blob/main/lib/jxl/render_pipeline/stage_blending.cc#L93-L98

Commenting this check out of `libjxl` makes it magically work...

Likewise, removing `color_encoding.transfer_function = JXL_TRANSFER_FUNCTION_LINEAR;` makes it load, but it's not linearized anymore 😅 

I'm having a little difficulty accessing any of these hidden `libjxl` variables from user code, so I can't even check if it's going to trip this...  `JxlDecoder` is an incomplete struct as far as the public API is concerned... Adding `#include <lib/jxl/decode.cc>` lets me check `if (!decoder.get()->metadata.m.xyb_encoded) {`, but this seems to cause cascading dependency issues during compilation, so that's no good...

I might be able to redefine the struct and cast it to change the value... or fork `libjxl` to remove the mysterious crash... but both seem beyond reasonable...

So for now, I'm leaving it so the CMS only activates when it's not sRGB, and just uses `toLinear()` when it is (since the breaking image _does_ technically report itself as sRGB).

Perhaps there will be a counter example in the conformance repo 😅

zalo:
Yep, there are a bunch of whacky edge cases in the conformance repo...
- The premultiplied/nonpremultiplied/alpha_triangle/patches images have odd brightness scaling; presumably because they're 9-bit or 12-bit 💀  (Microsoft's JPEGXL plugin for Photos doesn't render these correctly either)
- All of the files suffixed with `_5` aren't recognized as JPEG XL's; not sure what's up there...
- `spot` is really interesting, because it's the guts of `libjxl` that return `More than 4 channels not supported`... Even when I don't try to use the CMS... Will have to make my own testcase image there...

Everything else broadly works... better than I'd feared 😅

Tom94:
That sounds great already, thank you!

I’d actually be on board with forking libjxl to remove the crash case if it means we don’t have to worry about special cases on tev’s end. The faulty check seems to pertain to a niche color space anyway.

That said, maybe older revisions of jxl work just fine and this is a recent bug. Worth trying out older release tags, I think.

Another benefit of forking would be that we could relicense under GPL and then enable fast_float within jxl’s lcms2. In tev, this was a 2-4x speedup when loading HEIC images. (Though, without having benchmarked, I suppose I’m just hypothesizing a reason here.)

Tom94:
Heya, I spent a couple hours (more time than I was hoping) trying to get colors correct... jxl's built-in color conversion is either shockingly bad or I've been doing something very wrong. 😅

Two notable things:
- Half of the sRGB test images have completely wrong colors and brightness when using the built-in CMS. I suspect this is the problem that you thought was caused by bit depth. But no.  When I use external color processing, it's all fine. No brightness scaling required.
- Images with premultiplied alpha seem to have alpha multiplied *after* transferring into non-linear colors. What?!? I'm not sure if this is just a bug in the test images, but if this is a genuine thing in the jxl spec, that's terrible and defeats many reasons why one would use premultiplied alpha in the first place. Argh. Besides this rant, it implies we need to unmultiply alpha before doing all the color conversions. But! And here comes the kicker! When using jxl's built-in CMS, it *doesn't do the unmultiplication first*.

Long story short: I pushed a change to *only* use jxl's built-in color processing when opening an opsin image (XYB colorspace), because that'd be troublesome to handle ourselves. All other images (including sRGB) now use lcms2 directly (*after* unmultiplying alpha 🤦‍♂️).

A subset of the sRGB images now have slightly different brightness compared to their reference image counterparts (which wasn't present when manually undoing the sRGB transfer function) -- as far as I can see, this is because the reference png images have ICC profiles with corresponding brightness shifts attached... and tev doesn't support ICC profiles embedded in png images. Hence the difference.

Tom94:
All in all, I feel like the PR is in a good enough state to be merged. Wouldn't expect you to invest more time unless you really feel like it.

Remaining rough edges that I could also take care of if you prefer:
- two test images don't load, because output buffer size can't be read (?)
- the distinction between `num_color_channels` and `num_extra_channels` is currently ignored for the purposes of color conversion, which could lead to breakage in situations like 2+2 channels without alpha or >4 channels

Tom94:
Heads up that I've cleaned up the history a little bit to make every commit compile and to get rid of most of the back-and-forth testing that happened

Tom94:
Thanks again for all your work! I took care of the remaining things and cleaned up a bit. All the test images load now, including extra channels. Saving premultiplier linear sRGB also works now. 🎉

The only remaining unsupported things I can see that we miiiight want to support eventually is CMYK handling (which is super jank in jxl, requiring a special extra channel dance) and re-orientation of images based on metadata.

zalo:
Thank you so much!  This is awesome!

Sorry for dropping off for a day; work came up 😅

I’m excited for the potential!  tev is the only image viewer I know of that supports most of the features of JPEG-XL, so it will be invaluable for inspecting and comparing them so people can get comfortable with the format 😄

(And they’re generally a LOT smaller than .exr’s, which is what typically limited my usage of HDR environment maps, floating point textures, etc.)

Perhaps there are graphics folks who will want to use tev’s processing infrastructure as a reference for making .jxl texture loaders for more ubiquitous HDR 🧐

My personal use-case this time was,because I’m storing gigabytes of high precision ultrasound data, and .jxl gives me the best speed/compression tradeoff, but no other viewer could handle the range of the data 😄 

Thank you again!

PR #341, opened 2025-09-01
Flatpak

bash:
I am working on [packaging tev for Flatpak](https://codeberg.org/tautropfli/io.github.tom94.tev) 🚀 

The changes in this PR are currently applied as patches to the Flatpak,
but I would love to upstream them :)

WIP as I might find more things to patch ^^


bash:
I think this should be all the changes that I would like to upstream as part of packaging tev for Flatpak :)

Two questions are still open for me: 

## App ID

Flatpak apps use an [application ID](https://docs.flathub.org/docs/for-app-authors/requirements#application-id) in the reverse-DNS format. To get the app verified, the application ID should use a registered domain and place a special TXT file at a well-known path. There is special handling for [source code hosting websites](https://docs.flathub.org/docs/for-app-authors/verification#source-code-hosting-site) such as GitHub where one can use `io.github.owner.repository`.

I thus chose `io.github.tom94.tev`. Let me know if you prefer a different app ID (i.e. pointing to your website).

Once the app is submitted on Flathub (I will ask them to add you as a maintainer during submission) you can sign into Flathub using your GitHub account to [verify](https://docs.flathub.org/docs/for-app-authors/verification#im-publishing-an-app-on-flathub-how-do-i-get-it-verified) tev (if you want to of course ^^).

## Screenshot
I have tried to recreate the screenshot found in the README of this repository, but the two beach ball images don't overlap fully. Is there a way to move the data / reference windows around?

This is how far I got:
image

<sub>I'm not a regular user of tev myself, so this might be a silly question. But my friend uses tev (and loves it!)</sub>

bash:
> For the screenshot, I don't think one can replicate the one from the README these days. For one, the false color palette has changed (which is why the image in your screenshot is orange rather than purple). Plus, the beachballs only used to overlap because tev didn't handle data vs. display windows back in the day.

Aaah, that makes sense, thanks :)

> Is using the screenshot from the README an option?

Unfortunately not, since the metadata guidelines say that it should be up-to-date and take on Linux :) I will open a follow-up PR to add said screenshot.

PR #363, opened 2025-10-21
feat(ui): if possible, cap window resize at current monitor bounds

### LINKED ISSUE #362: [Wayland] The window is resized to a size greater than the screen resolution upon image load
Once an image is loaded, the app's window size changes drastically to (I assume) fit the contents of the image on the window. Unfortunately, that means the window might become too big for its actual display.

I went and analyzed the code, which led me to lines 1196-1202 in `ImageViewer.cpp`:

```cpp
// First image got added, let's select it.
if ((index == 0 && mImages.size() == 1) || shallSelect) {
    selectImage(image);
    if (!isMaximized()) {
        resizeToFit(sizeToFitImage(image)); // ← this line causes the issue, removing it keeps the window size as it was
    }
}
```

I verified 100% and 200% scaling factors in KDE Plasma 6.4.5 (Wayland) result in the same thing.

Is this intentional? If yes, would it make sense to limit that resizing behavior to the maximum logical screen dimensions, and only when the autofit option is set to off (since when it's on there doesn't seem to be a reason to resize the window)? If not, what would be the best way to resolve it?

Here's my system info:

- Operating System: Arch Linux 
- KDE Plasma Version: 6.4.5
- KDE Frameworks Version: 6.19.0
- Qt Version: 6.10.0
- Kernel Version: 6.17.3-arch2-1 (64-bit)
- Graphics Platform: Wayland
- Processors: 16 × AMD Ryzen 7 5800X 8-Core Processor
- Memory: 64 GiB of RAM (62.7 GiB usable)
- Graphics Processor: AMD Radeon RX 6600 XT

Displays:

- 3840x2160@160Hz, 200% scaling
- 1920x1080@165Hz, 100% scaling

Tom94:
Currently, obtaining the current monitor bounds is limited to

- [x] Wayland
- [x] X11
- [x] Windows
- [x] macOS

Fixes #362

### DIFF
--- a/src/ImageViewer.cpp
+++ b/src/ImageViewer.cpp
@@ -52,7 +52,10 @@ static const float CROP_MIN_SIZE = 3;
 ImageViewer::ImageViewer(
     const Vector2i& size, const shared_ptr<BackgroundImagesLoader>& imagesLoader, const shared_ptr<Ipc>& ipc, bool maximize, bool showUi, bool floatBuffer
 ) :
-    nanogui::Screen{size, "tev", true, maximize, false, true, true, floatBuffer}, mImagesLoader{imagesLoader}, mIpc{ipc} {
+    nanogui::Screen{size, "tev", true, maximize, false, true, true, floatBuffer},
+    mImagesLoader{imagesLoader},
+    mIpc{ipc},
+    mMaximizedLaunch{maximize} {
 
     auto tf = ituth273::fromWpTransfer(glfwGetWindowTransfer(m_glfw_window));
     mSupportsHdr = m_float_buffer || tf == ituth273::ETransferCharacteristics::PQ || tf == ituth273::ETransferCharacteristics::HLG;
@@ -69,7 +72,8 @@ ImageViewer::ImageViewer(
     // At this point we no longer need the standalone console (if it exists).
     toggleConsole();
 
-    // Get monitor configuration to figure out how large the tev window may maximally become.
+    // Get monitor configuration to figure out how large the tev window may maximally become. This will later get overwritten once
+    // glfwGetWindowCurrentMonitor() works (it does not while the window is still getting constructed).
     {
         int monitorCount;
         auto** monitors = glfwGetMonitors(&monitorCount);
@@ -84,9 +88,13 @@ ImageViewer::ImageViewer(
                 monitorMax = max(monitorMax, pos + size);
             }
 
-            mWorkAreaSize = min(mWorkAreaSize, max(monitorMax - monitorMin, nanogui::Vector2i{1024, 800}));
+            mMaxWindowSize = min(mMaxWindowSize, max(monitorMax - monitorMin, nanogui::Vector2i{1024, 800}));
         }
     }
+
+    // Try to get the current monitor size right away. Better to have it early. The function will get called again before every draw to
+    // handle monitor changes as well as the case where glfwGetWindowCurrentMonitor() did not work yet.
+    updateCurrentMonitorSize();
 
     m_background = Color{0.23f, 1.0f};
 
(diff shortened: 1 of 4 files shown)


src/ImageViewer.cpp
```diff
@@ -1713,7 +1713,21 @@ void ImageViewer::resizeToFit(nanogui::Vector2i targetSize) {
...
     // For sanity, don't make us larger than 8192x8192 to ensure that we don't break any texture size limitations of the user's GPU.
-    auto maxSize = mMaxSize;
+    auto maxSize = Vector2i{8192, 8192};
+
+    if (GLFWmonitor* monitor = glfwGetWindowCurrentMonitor(m_glfw_window); monitor != nullptr) {
+        Vector2i pos, size;
+        glfwGetMonitorWorkarea(monitor, &pos.x(), &pos.y(), &size.x(), &size.y());
+        if (glfwGetPlatform() == GLFW_PLATFORM_WAYLAND) {
+            size /= pixel_ratio();
+        }
```
fililip: Oh, sorry, I just compiled this branch and it seems that it's unnecessary now, and it actually reduces the max window size to an incorrect value. Removing these lines works correctly on both 1x and 2x scaling

fililip: Oh, that's because `glfwGetMonitorWorkarea` seems to return 1080p for my 4K monitor now... weird

Edit: only for 2x scaling, lower factors (but greater than 1) always return 2160p.

fililip: I think it's a different thing; `pixel_ratio()` is always correct (1, 1.5, 1.75, 2), but `glfwGetMonitorWorkarea` only returns `[1920, 1080]` for 200% scaling.

That division operation was definitely correct (it certainly works with fractional scaling), something seems to be wrong with the way GLFW interprets monitor sizes now. (Did a dependency update change something here?)

Tom94: At the place where `glfwGetMonitorWorkarea` was previously called, `pixel_ratio()` would always return 1 in the scenario I described above. My suspicion is that KDE sends over correspondingly down-scaled monitor work areas as it sends the correct content scale such that their product always reproduces your true resolution.

If you're up for testing, you could try printing the results of `glfwGetMonitorWorkarea` and `pixel_ratio()` at the top of the `ImageViewer` constructor.

(No dependency update -- I'm already maintaining forks of all the relevant dependencies, so this was an isolated change.)

fililip: I rebased and recompiled the branch, here are my results:

Constructor:

```
20:00:26 INFO     monitor 0 size: [1462999040, 32765]
20:00:26 INFO     monitor 1 size: [3840, 2160]
20:00:26 INFO     pixel_ratio(): 1
```


Primary screen:

```
20:01:06 INFO     monitor 0 size: [-184148240, 32764]
20:01:06 INFO     monitor 1 size: [3840, 2160]
20:01:06 INFO     pixel_ratio(): 2
20:01:06 INFO     current monitor size: [3840, 2160]
```

Secondary screen:

```
20:01:15 INFO     monitor 0 size: [1616729504, 32767]
20:01:15 INFO     monitor 1 size: [3840, 2160]
20:01:15 INFO     pixel_ratio(): 1
20:01:15 INFO     current monitor size: [1920, 1080]
```

This means that now the current monitor call returns the correct screen, but `glfwGetMonitors` appears to add an uninitialized(?) handle as the first element of that pointer array. Accessing the third element yields the secondary screen's dimensions:

```
20:03:45 INFO     monitor 0 size: [-708719424, 32764]
20:03:45 INFO     monitor 1 size: [3840, 2160]
20:03:45 INFO     monitor 2 size: [1920, 1080]
```

Tom94: Thank you. I'm honestly not sure what happened there. I can't reproduce the issue on my single monitor setup, but I tried another small change as a shot in the dark. 🤞

PR #370, opened 2025-11-22
Add release date to MetaInfo

tautropfli:
Follow up to #360 where I forgot to add the release date, failing the appstream validation 🙈


Tom94:
I just deleted the v1.19-qoi tag which should simplify things. Just using `v*.*` will work (also matches `v*.*.*`).

Also: is there a way to sort the tags naturally rather than alphabetically? (i.e. v1.20 should come after v1.3) Alternatively, sorting by date would also do.

Tom94:
What about getting rid of the patch file and adding the `TEV_CREATE_METAINFO` flag, though? I'd have suspected that this needs to be synced with the next release to avoid the flatpak repo being in a broken state.

PR #372, opened 2025-11-22
feat(imageviewer): add white level override setting

### LINKED ISSUE #369: Windows: add option to make HDR output not relative to the "SDR content brightness" slider
no idea why it is like that for HDR10 (PQ + BT.2020 primaries) pngs that are described as absolute while scRGB jxls are unaffected by the slider. I know chromium does it like that for pngs and I think it's horrible design having to set the SDR content brightness slider to value 31 to get accurate output when it's the "SDR" content brightness slider...

Tom94:
Adds a setting to override the automatically detected display white level.

Useful for displaying images with known white level (e.g. 203 nits for PQ / HLG HDR images) at absolute brightness instead of at the system's configured relative brightness level. 

**IMPORTANT**: this only works correctly when using a well-calibrated HDR display & when the OS/Compositor has HDR enabled.

In the future, **tev** could try detecting images whose encoding has a well-standardized white level to make absolute brightness automatic. Three roadblocks, though:
- There is currently no way for **tev** to know that it is running in a well calibrated HDR environment where adjusting the white level does something meaningful other than confuse the user (because their image no longer matches what e.g. Chrome looks like).
- A few of the PQ/HLG test images that I have lying around assume a relative rather than an absolute white level... so, there seems to be a precedent for ignoring the standard. Ergo: absolute white levels can't be made the default, even if feasible.
- Many HDR image formats (e.g. EXR or JXL when not relying on PQ/HLG transfers) don't have standardized white level metadata at all, so this would be a PQ/HLG exclusive feature as far as I can tell.

Fixes #369

### DIFF
--- a/include/tev/ImageCanvas.h
+++ b/include/tev/ImageCanvas.h
@@ -91,14 +91,17 @@ public:
     void setCrop(const std::optional<Box2i>& crop) { mCrop = crop; }
     Box2i cropInImageCoords() const;
 
-    auto backgroundColor() { return mShader->backgroundColor(); }
-    void setBackgroundColor(const nanogui::Color& color) { mShader->setBackgroundColor(color); }
-
     void fitImageToScreen(const Image& image);
     void resetTransform();
 
+    std::optional<float> whiteLevelOverride() const { return mWhiteLevelOverride; }
+    void setWhiteLevelOverride(std::optional<float> value) { mWhiteLevelOverride = value; }
+
     bool clipToLdr() const { return mClipToLdr; }
     void setClipToLdr(bool value) { mClipToLdr = value; }
+
+    auto backgroundColor() { return mBackgroundColor; }
+    void setBackgroundColor(const nanogui::Color& color) { mBackgroundColor = color; }
 
     EInterpolationMode minFilter() const { return mMinFilter; }
     void setMinFilter(EInterpolationMode value) { mMinFilter = value; }
@@ -151,7 +154,9 @@ private:
     float mOffset = 0;
     float mGamma = 2.2f;
 
+    std::optional<float> mWhiteLevelOverride = std::nullopt;
     bool mClipToLdr = false;
+    nanogui::Color mBackgroundColor = nanogui::Color(0, 0, 0, 0);
 
     EInterpolationMode mMinFilter = EInterpolationMode::Trilinear;
     EInterpolationMode mMagFilter = EInterpolationMode::Nearest;
--- a/include/tev/ImageViewer.h
+++ b/include/tev/ImageViewer.h
@@ -121,12 +121,13 @@ public:
     void setMagFilter(EInterpolationMode value) { mImageCanvas->setMagFilter(value); }
 
     ETonemap tonemap() const { return mImageCanvas->tonemap(); }
-
     void setTonemap(ETonemap tonemap);
 
     EMetric metric() const { return mImageCanvas->metric(); }
-
     void setMetric(EMetric metric);
+
+    std::optional<float> overridingWhiteLevel() const;
+    void setOverridingWhiteLevel(std::optional<float> value);
 
     nanogui::Vector2i sizeToFitImage(const std::shared_ptr<Image>& image);
     nanogui::Vector2i sizeToFitAllImages();
@@ -211,26 +212,26 @@ private:
     bool mRequiresFilterUpdate = true;
     bool mRequiresLayoutUpdate = true;
 
-    nanogui::Widget* mVerticalScreenSplit;
+    nanogui::Widget* mVerticalScreenSplit = nullptr;
 
-    nanogui::Widget* mSidebar;
-    nanogui::Button* mHelpButton;
-    nanogui::Widget* mSidebarLayout;
+    nanogui::Widget* mSidebar = nullptr;
+    nanogui::Button* mHelpButton = nullptr;
+    nanogui::Widget* mSidebarLayout = nullptr;
 
-    nanogui::Widget* mFooter;
+    nanogui::Widget* mFooter = nullptr;
     bool mShouldFooterBeVisible = false;
 
-    nanogui::Label* mExposureLabel;
-    nanogui::Slider* mExposureSlider;
+    nanogui::Label* mExposureLabel = nullptr;
+    nanogui::Slider* mExposureSlider = nullptr;
 
-    nanogui::Label* mOffsetLabel;
-    nanogui::Slider* mOffsetSlider;
+    nanogui::Label* mOffsetLabel = nullptr;
+    nanogui::Slider* mOffsetSlider = nullptr;
 
-    nanogui::Label* mGammaLabel;
-    nanogui::Slider* mGammaSlider;
+    nanogui::Label* mGammaLabel = nullptr;
+    nanogui::Slider* mGammaSlider = nullptr;
 
-    nanogui::Widget* mTonemapButtonContainer;
-    nanogui::Widget* mMetricButtonContainer;
+    nanogui::Widget* mTonemapButtonContainer = nullptr;
+    nanogui::Widget* mMetricButtonContainer = nullptr;
 
     std::shared_ptr<BackgroundImagesLoader> mImagesLoader;
     std::shared_ptr<Ipc> mIpc;
@@ -240,36 +241,36 @@ private:
 
     std::vector<std::shared_ptr<Image>> mImages;
 
-    MultiGraph* mHistogram;
+    MultiGraph* mHistogram = nullptr;
     std::set<std::shared_ptr<Image>> mToBump;
 
-    nanogui::TextBox* mFilter;
-    nanogui::Button* mRegexButton;
+    nanogui::TextBox* mFilter = nullptr;
+    nanogui::Button* mRegexButton = nullptr;
 
-    nanogui::Button* mWatchFilesForChangesButton;
+    nanogui::Button* mWatchFilesForChangesButton = nullptr;
     std::chrono::steady_clock::time_point mLastFileChangesCheckTime = {};
 
-    nanogui::Button* mAutoFitToScreenButton;
+    nanogui::Button* mAutoFitToScreenButton = nullptr;
 
     // Buttons which require a current image to be meaningful.
     std::vector<nanogui::Button*> mCurrentImageButtons;
-    nanogui::Button* mImageInfoButton;
+    nanogui::Button* mImageInfoButton = nullptr;
     ImageInfoWindow* mImageInfoWindow = nullptr;
 
     // Buttons which require at least one image to be meaningful
     std::vector<nanogui::Button*> mAnyImageButtons;
 
-    nanogui::Button* mPlayButton;
-    nanogui::IntBox<int>* mFpsTextBox;
+    nanogui::Button* mPlayButton = nullptr;
+    nanogui::IntBox<int>* mFpsTextBox = nullptr;
     std::chrono::steady_clock::time_point mLastPlaybackFrameTime = {};
 
-    nanogui::Widget* mImageButtonContainer;
-    nanogui::Widget* mScrollContent;
-    nanogui::VScrollPanel* mImageScrollContainer;
+    nanogui::Widget* mImageButtonContainer = nullptr;
+    nanogui::Widget* mScrollContent = nullptr;
+    nanogui::VScrollPanel* mImageScrollContainer = nullptr;
 
-    ImageCanvas* mImageCanvas;
+    ImageCanvas* mImageCanvas = nullptr;
 
-    nanogui::Widget* mGroupButtonContainer;
+    nanogui::Widget* mGroupButtonContainer = nullptr;
     std::string mCurrentGroup;
 
     HelpWindow* mHelpWindow = nullptr;
@@ -288,8 +289,11 @@ private:
 
     size_t mClipboardIndex = 0;
 
+    // HDR support
     bool mSupportsHdr = false;
-    nanogui::Button* mClipToLdrButton;
+    nanogui::Button* mClipToLdrButton = nullptr;
+    nanogui::FloatBox<float>* mWhiteLevelBox = nullptr;
+    nanogui::Button* mWhiteLevelOverrideButton = nullptr;
 
     int mDidFitToImage = 0;
 
--- a/include/tev/UberShader.h
+++ b/include/tev/UberShader.h
@@ -48,18 +48,16 @@ public:
         float exposure,
         float offset,
         float gamma,
+        float colorMultiplier,
         bool clipToLdr,
+        const nanogui::Color& backgroundColor,
         ETonemap tonemap,
         EMetric metric,
         const std::optional<Box2i>& crop
     );
 
-    const nanogui::Color& backgroundColor() { return mBackgroundColor; }
-
-    void setBackgroundColor(const nanogui::Color& color) { mBackgroundColor = color; }
-
 private:
-    void bindCheckerboardData(const nanogui::Vector2f& pixelSize, const nanogui::Vector2f& checkerSize);
+    void bindCheckerboardData(const nanogui::Vector2f& pixelSize, const nanogui::Vector2f& checkerSize, const nanogui::Color& backgroundColor);
 
     void bindImageData(
         nanogui::Texture* textureImage, const nanogui::Matrix3f& transformImage, float exposure, float offset, float gamma, ETonemap tonemap
@@ -70,8 +68,6 @@ private:
     nanogui::ref<nanogui::Shader> mShader;
     nanogui::ref<nanogui::Texture> mColorMap;
     nanogui::ref<nanogui::Texture> mDitherMatrix;
-
-    nanogui::Color mBackgroundColor = nanogui::Color(0, 0, 0, 0);
 };
 
 } // namespace tev
--- a/src/ImageCanvas.cpp
+++ b/src/ImageCanvas.cpp
@@ -103,7 +103,9 @@ void ImageCanvas::draw_contents() {
         mExposure,
         mOffset,
         mGamma,
+        mWhiteLevelOverride ? (*mWhiteLevelOverride / glfwGetWindowSdrWhiteLevel(glfwWindow)) : 1.0f,
         mClipToLdr,
+        mBackgroundColor,
         mTonemap,
         mMetric,
         imageSpaceCrop
--- a/src/ImageViewer.cpp
+++ b/src/ImageViewer.cpp
@@ -187,33 +187,94 @@ ImageViewer::ImageViewer(
             return button;
         };
 
-        mCurrentImageButtons.push_back(makeButton("Normalize", [this]() { normalizeExposureAndOffset(); }, 0, "Shortcut: N"));
+        mCurrentImageButtons.push_back(makeButton(
+            "Norm.",
+            [this]() { normalizeExposureAndOffset(); },
+            0,
+            "Normalize image such that the smallest pixel value is displayed as 0 and the largest as 1.\n\n"
+            "Shortcut: N"
+        ));
         makeButton("Reset", [this]() { resetImage(); }, 0, "Shortcut: R");
 
         if (mSupportsHdr) {
-            mClipToLdrButton = new Button{buttonContainer, "LDR", 0};
-            mClipToLdrButton->set_font_size(15);
-            mClipToLdrButton->set_change_callback([this](bool value) { mImageCanvas->setClipToLdr(value); });
-            mClipToLdrButton->set_tooltip(
-                "Clips the image to [0,1] as if displayed on a low dynamic range (LDR) screen.\n\n"
-                "Shortcut: U"
-            );
-            mClipToLdrButton->set_flags(Button::ToggleButton);
+            auto hdrPopupButton = new PopupButton{buttonContainer, "HDR", 0};
+            hdrPopupButton->set_font_size(15);
+            hdrPopupButton->set_chevron_icon(0);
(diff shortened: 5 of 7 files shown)


EndlesslyFlowering:
> A few of the PQ/HLG test images that I have lying around assume a relative rather than an absolute white level... so, there seems to be a precedent for ignoring the standard. Ergo: absolute white levels can't be made the default, even if feasible.

is this about the "rendering intent" flag?

EndlesslyFlowering:
The 203 nits value is a false narrative for PQ and any other scene referred format imho. It's only valid for HLG where it's somewhat expected (not standardised) that the diffuse/reference white point gets mapped to 75% HLG. And even then only at 1000 nits target max output it's 203 nits. Also I would argue this only applies to TV programming (like news or live TV) where the the diffuse white level is kept somewhat constant. For any other HDR content diffuse white level is scene dependent. Afaik the average brightness of an image is defined to be about equal to the mid grey value. And since the mid grey is expected to be at 18% reflectance you can derive diffuse white easily since it's at 100% reflectance.
The whole thing about 203 nits being a "standard" started in BT.2408 which is not even an ITU recommendation and "only" a report. There are "competing" values from Dolby and some other entity that I cannot remember atm, which are 140 and 120 nits respectively. I say this as someone who pushed the 203 value before in the past. My issue is that people get too stuck on arbitrary numbers like the 203 nits number and it basically blindly gets accepted as some ground truth when it's "only" the ITU doing some test recordings. I am not saying this to put the ITU down or anything, the test recordings that let to the 203 nits value make sense in their context but not everything is that context.

Tom94:
I understand the desire to desire for HDR10 and scRGB (more generally scene referred formats) to display at absolute brightness, but I feel like we're back at square one now: I don't know how to reliably detect those. Do you have a suggestion?

It can't be just the color spaces. E.g. sRGB primaries w/ linear transfer have been around for ages now, so there has to be some additional piece of metadata that identifies an image as scRGB. Same issue with HDR10: just because an image uses BT.2020 primaries w/ PQ transfer does not make it scene referred. I thought the PQ transfer alone would imply scene referredness (which is why I talked about PQ so much above), but you made multiple fair points that that's not necessarily the case (e.g. the ones on UltraHDR).

EndlesslyFlowering:
> I understand the desire to desire for HDR10 and scRGB (more generally scene referred formats) to display at absolute brightness, but I feel like we're back at square one now: I don't know how to reliably detect those. Do you have a suggestion?
> 
> It can't be just the color spaces. E.g. sRGB primaries w/ linear transfer have been around for ages now, so there has to be some additional piece of metadata that identifies an image as scRGB. Same issue with HDR10: just because an image uses BT.2020 primaries w/ PQ transfer does not make it scene referred. I thought the PQ transfer alone would imply scene referredness (which is why I talked about PQ so much above), but you made multiple fair points that that's not necessarily the case (e.g. the ones on UltraHDR).

sRGB with linear transfer is essentially always scRGB, the standard was just never revised for floating point afaik. it's standardised for an int16 and int12 format but not for floating point, there is only a single Microsoft page about DirectX that mentions how it works in fp. scRGB has a few names: `DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709` in DirectX, `COLOR_SPACE_EXTENDED_SRGB_LINEAR_EXT` in Vulkan (there is also `COLOR_SPACE_EXTENDED_SRGB_NON_LINEAR_EXT`, which also allows values above 1 and below 0), [`extendedLinearSRGB`](https://developer.apple.com/documentation/coregraphics/cgcolorspace/extendedlinearsrgb) for Apple, internally in chromium they also call it similarly iirc, in jxl it's `Color space: RGB, D65, sRGB primaries, Linear transfer function` and the shorthand is `RGB_D65_SRG_Rel_Lin` (you can also just have an image that uses an integer format and tag it with that and it still would be correct, so in this case it covers both linear sRGB and scRGB depending on the underlying data type being used).
scRGB is just sRGB primaries with a linear transfer and values above 1 and below 0 are allowed. that was always the point as I understand it. a wide gamut and HDR format that is easily compatible with sRGB.
And yes scRGB has existed since 2003. it's standardised under "IEC 61966-2-2" (only the int16 and int12 version afaik).

about HDR10 and PQ: PQ is 100% scene referred. The PQ EOTF is standardised to output nits:
see the SMPTE standard [here](https://pub.smpte.org/doc/st2084/):
image
and the latest ITU recommendation [here](https://www.itu.int/rec/R-REC-BT.2100):
image

the point I was making about UltraHDR is that it seems like it separates itself format even though the input can be scene referred. I have read the code a bit [here](https://github.com/google/libultrahdr/blob/5ed39d67cd31d254e84ebf76b03d4b7bcc12e2f7/lib/src/gpu/applygainmap_gl.cpp#L245-L314) now and it looks like it when the input image when creating an UltraHDR image is PQ it will also output PQ when decoding it. As in the format remembers what the input originally was. So I was wrong about that, sorry. I thought when applying the gain map that it would always be the same algorithm and it was trying to be separate from the known standards (as google likes to do).

about detection: shouldn't all image formats provide you with the necessary metadata? on Windows there is an unwritten rule that an untagged fp image is scRGB and Linux inherits that. similarly how an untagged SDR image is assumed to be sRGB. like tev outputs fp16 in OpenGL and it only works because of that unwritten rule as OpenGL has no colour space option. similarly you can set the swapchain in DirectX to fp16 and just not set the colour space and it will be read as scRGB (not like there is any other colour space allowed for a fp16 swapchain).

EndlesslyFlowering:
> Only the 16 bit integer variant. The 12-bit and 14-bit variants use the sRGB transfer. Which already makes the waters murky. But more than that, the 2003 standard is still SDR compatible, albeit wide gamut[1], so should scale with SDR brightness controls as far as I can tell. I don't think scene-referredness is implied here. (Or at least not beyond how sRGB's 80 nits encoding viewing environment has been widely ignored anyway.)

there is 14 bit variant? the 12 bit one is named scRGB-nl with nl standing for non linear. The scRGB spec was made before scene referred even existed I think so it makes sense that it is described to be relatively.

> This is my actual point. People have been using linear floating point w/ sRGB primaries for HDR content for ages, way before Microsoft adopted it and decided to display it as scene-referred when in HDR mode. But the waters get murkier still, because when Windows is set so SDR mode [the same linear fp16 sRGB buffers get interpreted as display-referred](https://learn.microsoft.com/en-us/windows/win32/direct3darticles/high-dynamic-range#system-composition-using-a-high-bit-depth-canonical-color-space-1) and are still called scRGB in their docs.

this happens with HDR10 too as it gets mapped into scRGB because that is what the dwm uses as intermediate colour space. scRGB is also used to output WCG in SDR though that stuff is really complicated to make it work. scRGB or any colour space the dwm accepts gets force converted to sRGB too when outputting SDR. scRGB is weird as it's only meant as intermediate colour space but Microsoft being Microsoft also forced it on other things.

> You mention macOS as having the extendedLinearSRGB setting. tev was actually using it and... it's always interpreted as display-referred. (Currently using a different variant, but same behavior.) There is no macOS system setting or color space that you can choose to render scene-referred scRGB. You actually have to jump through quite a few hoops to configure an app to render absolute rather than relative brightnesses. It's like you say: this is for backwards compatibility with SDR sRGB, but it goes directly contrary to being scene-referred. I think it's no coincidence that all the constants for configuring linear sRGB color spaces (DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709, COLOR_SPACE_EXTENDED_SRGB_LINEAR_EXT, extendedLinearSRGB) say "SRGB" and not "SCRGB".

well would you know it the same would have happened if you had used HDR10 as colour space. I know this information only from 2nd hand, so you would need to do the testing yourself to verify that. `extendedLinearSRGB` behaves exactly how scRGB fp is defined. for example you could encode a scene referred image as HDR10 and tell macOS that it's HDR10 and then use the same image and encode it as scRGB and tell macOS that it's `extendedLinearSRGB` and the image would look the exact same while both are also adjustable by the global brightness slider. this design is intentional from what I understand and works because macOS seems to be fully colour managed and they have control over the whole system usually while Windows is not (there is no global brightness slider unless you are on a laptop). Apple's XDR display can be made to be scene referred but according to them you should only do that if you have a reference viewing environment. Windows again has nothing similar.
`DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709`, `COLOR_SPACE_EXTENDED_SRGB_LINEAR_EXT` and `extendedLinearSRGB` all behave like scRGB fp in their described contexts while they can also work as screen referred in SDR or EDR if you want to. similarly you could use HDR10 to output screen referred if you really want to on Windows and macOS and maybe even Linux. amazing right?

> Unfortunately not. I don't know a single image format that tells me unambiguously that it is an fp16 scRGB image (therefore scene-referred) rather than just a linear fp16 image w/ sRGB primaries that tev should render backwards-compatible with SDR sRGB. Like 99% of the EXR images I've got lying around, for example.
>
> I wish it were as simple as taking the image data and rendering it to an FP16 buffer without any processing. If I do this, I either break SDR compatibility (on Windows in HDR mode) or scene-referredness (on Windows in SDR mode, on macOS always, and on Linux if I don't explicitly tell the compositor that I'm going to give it scRGB instead of linear sRGB).

my solution to that on Windows and Linux would be providing a slider in HDR that defaults to 80 nits as default white level which would match scRGB fp behaviour. on macOS nothing like that should be needed as the macOS users seem to be used to adjust their display brightness when the content is not at the level they want.

> (As an aside: some fp16 image formats don't tag color spaces at all. They just provide a color conversion matrix. For example, when a camera estimates the white point of a photo that it's taking and expresses its colors in whitepoint-corrected way. It would be silly to match such a matrix to sRGB primaries and change absolute/relative brightness behavior only if that matrix happens to match exactly.)

what I described above should only apply if there is no such metadata available. I assume the matrix is defined from some standardised neutral point? I assume you have to convert it to sRGB/BT.709 primaries referencing that matrix in some way, so that should not be an issue.

Tom94:
> similarly you could use HDR10 to output screen referred if you really want to on Windows and macOS and maybe even Linux. amazing right?

The problem was not how to output absolute brightness on supported systems -- I understand that process -- but how to reliably interpret how image authors intended their floating point linear sRGB images to be viewed. (I also mentioned above that it's difficult to reliably detect when **tev** *runs* on a system that supports absolute brightness, but I feel like I got that more of less in hand now.)

> for example you could encode a scene referred image as HDR10 and tell macOS that it's HDR10 and then use the same image and encode it as scRGB and tell macOS that it's extendedLinearSRGB and the image would look the exact same while both are also adjustable by the global brightness slider.

No. And this is really weird, actually. If I configure macOS with `kCGColorSpaceITUR_2020_PQ`/`kCGColorSpaceITUR_2100_PQ` the brightness doesn't match that of scRGB. But the conversion is also not 10000/203... it's somewhere in the middle. Roughly 10000/120. I played around with a few settings, monitor brightness, etc., but couldn't figure out what's going on here. Might be system specific -- I'm testing this on a macbook, so it doesn't have a reference mode like the Pro Display XDR.

That said, if I open an HDR10 image on macOS's default image viewer "Preview" (which is HDR capable), it matches **tev**'s current behavior exactly. I.e. it displays equivalently to mapping PQ to scRGB by dividing nits by 203. Same as the major browsers that you were complaining about earlier.

> my solution to that on Windows and Linux would be providing a slider in HDR that defaults to 80 nits as default white level which would match scRGB fp behaviour.

I was thinking along similar lines, though I cannot make the setting active by default for aforementioned OS/browser compatibility reasons. Plus, most of **tev**'s users are graphics researchers and renderer developers who for over a decade have produced linear sRGB EXR images that are expected to display with relative brightness.

Anyway, thanks for the feedback. Here's what I'll do: I'll expand the display white level UI from this PR with a best-effort "inferred image white level" that'll be 203 nits for PQ and HLG and 80 for all other images. There'll be a toggle / CLI option to automatically set the display white level to that inferred image white level.

Maybe also interesting: I scanned through all the image formats **tev** supports yesterday and found some metadata I had missed before.
- PNG, AVIF and HEIC can optionally carry maxFALL and maxCLL metadata -- if they do, those metadata will override the image white level accordingly. That said, those formats can't carry floating point data, so, at best, they anchor HLG to a scene and at worst contradict PQ which, at least in principle, already is scene-referred. I also haven't found any images in my test sets that actually carry this metadata, except for [Eric Chan's](https://people.csail.mit.edu/ericchan/hdr/) whose images never disappoint, so... `¯\_(ツ)_/¯`
- JXL images can carry something called an "intensity_target". Based on the docs, it'a a loose maxCLL (may be larger than the brightest pixel), so can't be used to get a concrete brightness, but... I'll make **tev** assume it's exact for the time being and see how it goes. As far as I can tell, this makes JXL the only image format that, at least in principle, allows tagging linear sRGB images with absolute brightness information.

> what I described above should only apply if there is no such metadata available. I assume the matrix is defined from some standardised neutral point? I assume you have to convert it to sRGB/BT.709 primaries referencing that matrix in some way, so that should not be an issue.

Nevermind, I was trying (and perhaps failing) to make a different point. Not worth going down that rabbit hole.

EndlesslyFlowering:
> No. And this is really weird, actually. If I configure macOS with kCGColorSpaceITUR_2020_PQ/kCGColorSpaceITUR_2100_PQ the brightness doesn't match that of scRGB. But the conversion is also not 10000/203... it's somewhere in the middle. Roughly 10000/120. I played around with a few settings, monitor brightness, etc., but couldn't figure out what's going on here. Might be system specific -- I'm testing this on a macbook, so it doesn't have a reference mode like the Pro Display XDR.

I guess the information is not correct then. They also had a macbook.

> That said, if I open an HDR10 image on macOS's default image viewer "Preview" (which is HDR capable), it matches tev's current behavior exactly. I.e. it displays equivalently to mapping PQ to scRGB by dividing nits by 203. Same as the major browsers that you were complaining about earlier.

🚮

> Anyway, thanks for the feedback. Here's what I'll do: I'll expand the display white level UI from this PR with a best-effort "inferred image white level" that'll be 203 nits for PQ and HLG, 100 for certain transfers (e.g. SDR BT.709, SMPTE ST 2080-1, and extensions) and 80 for all other images. There'll be a toggle / CLI option to automatically set the display white level to that inferred image white level.

Thanks you and thanks for listening me out. I'll check it out.

> PNG, AVIF and HEIC can optionally carry maxFALL and maxCLL metadata -- if they do, those metadata will override the image white level accordingly (== maxCLL / maxPixel). That said, those formats can't carry floating point data, so, at best, they anchor HLG to a scene and at worst contradict PQ which, at least in principle, already is scene-referred. I also haven't found any images in my test sets that actually carry this metadata, except for [Eric Chan's](https://people.csail.mit.edu/ericchan/hdr/) whose images never disappoint, so... ¯\_(ツ)_/¯

maxCLL and maxFALL are for tone mapping purposes, and get used when the image is brighter than the display's capabilities.

> JXL images can carry something called an "intensity_target". Based on the docs, it'a a loose maxCLL (may be larger than the brightest pixel), so can't be used to get a concrete brightness, but... I'll make tev assume it's exact for the time being and see how it goes. As far as I can tell, this makes JXL the only image format that, at least in principle, allows tagging linear sRGB images with absolute brightness information.

when using libjxl's cjxl to encode a png image encoded with HDR10 and tagged properly to jxl maxCLL gets mapped to "intensity_target". apparently "intensity_target" is a way that jxl adjusts its encoding to be more perceptually relevant: https://www.fractionalxperience.com/ux-ui-graphic-design-blog/why-jpeg-xl-ignoring-bit-depth-is-genius .

EndlesslyFlowering:
> The problem was not how to output absolute brightness on supported systems (where I don't count macOS as one) -- I understand that process -- but how to reliably interpret how image authors intended their floating point linear sRGB images to be viewed. I brought up that sRGB modes behave differently depending on context to illustrate why that's the case. (I also mentioned above that it's difficult to reliably detect when tev runs on a system that supports absolute brightness, but I feel like I got that more of less in hand now.)

that was a more a jab at how weirdly inconsistent everything is.

EndlesslyFlowering:
> I was hoping I could use it anyway, but turns out it's just too unreliable in practice, even when ignoring default values of 255. :(
> 
> For one, lossy compression destroys the maximum pixel's brightness when the image is even slightly noisy or peaky around it. So computing maxCLL/maxPixel is bust. But worse, even maxFALL (and maxCLL in losslessly encoded images) seem unreliable in most of the test images I have. Contradicting PQ by significant margins. I suspect that the images I have were extracted from videos whose maxCLL/maxFALL were kept.

maxCLL is defined at 95% of the actual max too. also maxCLL is not in nits but the max channel value with BT.2020 primaries instead of the luminance. it's still described as nits though...

EndlesslyFlowering:
ok so about `intensity_target` in jxl, I got some things wrong:
for absolute transfers it gets repurposed as max CLL and otherwise it describes what nits 1 equals to: https://github.com/libjxl/libjxl/blob/53042ec537712e0df08709524f4df097d42174bc/lib/include/jxl/cms_interface.h#L86-L99

Tom94:
> that was a more a jab at how weirdly inconsistent everything is.

Ah, yeah, I'm 100% with you on that. 😅

> maxCLL is defined at 95% of the actual max too. also maxCLL is not in nits but the max channel value with BT.2020 primaries instead of the luminance. it's still described as nits though...

Woot?! Do you know why? That seems so weird to me. I would understand if they defined it as the 95th percentile of pixels -- that'd at least make it resilient to noise and outliers -- but 95% of a component-max of BT.2020? Lum is only negligibly more expensive to compute since you'd have to look at all the channels to determine the max anyway... and it'd be nicely independent of the choice of primaries. Aaaaaargh.

I assume you got that straight from the standard? I didn't feel like paying (around) 100 dollars just to read the primary source, but I should really start doing that... there are [plenty](https://professionalsupport.dolby.com/s/article/Calculation-of-MaxFALL-and-MaxCLL-metadata?language=en_US) of [resources](https://www.linkedin.com/pulse/hdr-10-metadata-smpte-st2086-maxfall-maxcll-carlos-carmona) and [code](https://github.com/HDRWCG/HDRStaticMetadata/blob/master/hdrgenerator.cpp#L183) on the internet saying otherwise. At least the [PNG spec](https://www.w3.org/TR/png-3/#cLLI-chunk) mentions that there may be "algorithmic filters", but even they say it's cd/m^2.

Anyway, even if taking this into account, Eric Chan's maxCLL and maxFALL are unfortunately still wildly inconsistent. Sometimes off (relative to the max. and avg. pixel values) w.r.t. to each other by over 2x.

> ok so about intensity_target in jxl, I got some things wrong:
for absolute transfers it gets repurposed as max CLL and otherwise it describes what nits 1 equals to

According to their headers only if the image is unsigned int and not if it's float. That's why I hoped it could be used.
```cpp
  /** Upper bound on the intensity level present in the image in nits. For
   * unsigned integer pixel encodings, this is the brightness of the largest
   * representable value. The image does not necessarily contain a pixel
   * actually this bright. An encoder is allowed to set 255 for SDR images
   * without computing a histogram.
   * Leaving this set to its default of 0 lets libjxl choose a sensible default
   * value based on the color encoding.
   */
```
But... even [their own conformance test suite](https://github.com/libjxl/libjxl) doesn't follow this guideline in their (single) floating point test image.

Tom94:
> ahhhhh, sorry I totally misremembered that: it's 99.5% percentile for moving images and for still images it's 99.99%
[...]

Ah, that makes a lot more sense. And good point about tonemapping. I was thinking of the typical definitions of popular tonemapping functions in terms of luminance, but you're of course right that "luminance" needs to be interpreted creatively to be able to guarantee no clipping.

> this Microsoft HDR sample code also mentions that it's defined as the max of RGB in "CEA 861.3"

I quickly [tried](https://github.com/Tom94/tev/commit/255a4c0643492058e998fb1e4aa96b57481a8a21) 99.99th percentile and the values in my test images sadly still seem bogus, regardless of whether I use `max(RGB)` or actual luminance.

> just tested it [...] and it works using cjxl

Could you share the cjxl command you used as well as the image?

Tom94:
For what it's worth, while digging deeper into various resources and cleaning up **tev**'s handling of relative vs absolute rendering intents, I found that what we've been calling "scene referred" in this thread is very different from what's commonly understood as scene referred.

No need to reply -- I'm noting this down mostly for others who've been following this thread and for my future self.

---------------

HDR10, for one, seems to be widely considered display referred. (Because describes precisely how a display should render the content.)

HLG, on the other hand, seems to be widely considered scene referred. (Because it's treated as roughly representing scene light and leaving rendition mostly up to the display pipeline.)

SDR sRGB is also considered display referred (reference viewing environment with 80 nits, etc.). But some formats, like OpenEXR, also use sRGB primaries for representing scene-referred signals (typically linear or log-space). Often, like with PFM, the correct interpretation is unclear and depends on user (no metadata).

Notably, this is independent of whether brightness is relative or absolute. Just that for display-referred formats, absolute brightness represents how bright the image should be displayed, while for scene-referred formats it represents how bright the signal was when captured. And how to display it is left to the user/tool.

----------------

This rabbit hole mostly came up because I wanted more principled handling of rendering intents in **tev**'s color space conversions. And, when constrained to the colorimetrically accurate intents (`AbsoluteColorimetric` and `RelativeColorimetric`), it looks very much like the sensible choice is
- scene-referred <-> absolute
  - == convert w/o white point adaptation because data is independent of viewing conditions
- display-referred <-> relative
  - == convert w/ white point adaptation because data is already adapted to a display standard's white point

(Of course **tev** will let rendering intents from image metadata / ICC profiles override these defaults, if present.)

In practice, rendering intents won't change much and this is largely an academic point... because 99% of images share the D65 white point with display standards (sRGB, BT.2020, etc. all use D65). So it makes no difference whether white point adaptation is performed or not -- it's a no-op anyway.

PR #386, opened 2025-12-02
Bundle all sources together

tautropfli:
I think it would be nice to have a bundle of all sources (including submodules).
This moves downstream issues ([flatpak build](https://github.com/flathub/io.github.tom94.tev/pull/5)) upstream :)

### DIFF
--- a/.github/workflows/bundle-sources.yml
+++ b/.github/workflows/bundle-sources.yml
@@ -1,0 +1,28 @@
+name: Bundle Sources
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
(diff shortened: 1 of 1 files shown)


Tom94:
This feels hacky to me. A git clone with submodules is such a standard operation that it should be fine without an extra indirection.

I can freshly clone locally without issue -- and the nix package which also does a recursive clone behind the scenes builds fine as well. I also never heard of someone installing via AUR having a problem (recursive clone as well).

Let's wait a bit and try the flatpak build again later. If it stays broken, we can still revisit this, I suppose. 😅

Thanks for looking into it in any case!

PR #408, opened 2026-02-01
fix(window): resize movement on secondary monitor with scale

### LINKED ISSUE #407: Window Position Changes on Image Load / Refresh
This issue seems a bit more involved than my prior PRs, and is partially related to GLFW limitations, so I wanted to discuss it here first.

At least on Windows, if you have the tev window open on a separate monitor and try to load or refresh an image, it will move the tev window back to the primary monitor.

This is caused by ``ImageViewer::resizeToFit()``, where ``set_size()`` and ``move_window()`` cause GLFW to resize / move the window in the primary monitor. Part of the issue is that the window has a null monitor pointer, and GLFW still has no way to get the window's current monitor. I saw that this has been discussed before in #362. The cleanest solution is probably to fix glfw/glfw#1699 and update the logic accordingly.

I'm mainly opening the issue for visibility. I think this issue could be fixed by calling ``move_window()`` with the corresponding offset, and I'm happy to help with this fix if it's desired, although I don't quite understand the workarea clamping logic going on in that function. But maybe it's more efficient to just work on the GLFW issue...

Tom94:
If tev is on a non-primary monitor (with position != 0) that has non-1 scale (worse if scale is different from primary monitor), movement due to staying centered after resizing was wrong on Windows and X11. This commit fixes this.

Fixes #407

### DIFF
--- a/include/tev/ImageViewer.h
+++ b/include/tev/ImageViewer.h
@@ -355,6 +355,7 @@ private:
     // Misc state tracking variables
     int mDidFitToImage = 0;
 
+    nanogui::Vector2i mMinWindowPos = {0, 0};
     nanogui::Vector2i mMaxWindowSize = {8192, 8192};
 
     bool mInitialized = false;
--- a/include/tev/ImageViewer.h
+++ b/include/tev/ImageViewer.h
@@ -168,6 +168,9 @@ public:
     bool autoFitToScreen() const;
     void setAutoFitToScreen(bool value);
 
+    bool resizeWindowToFitImageOnLoad() const;
+    void setResizeWindowToFitImageOnLoad(bool value);
+
     void maximize();
     bool isMaximized();
     void toggleMaximized();
@@ -282,6 +285,7 @@ private:
     std::chrono::steady_clock::time_point mLastFileChangesCheckTime = {};
 
     nanogui::Button* mAutoFitToScreenButton = nullptr;
+    nanogui::Button* mResizeWindowToFitImageOnLoadButton = nullptr;
 
     // Buttons which require a current image to be meaningful.
     std::vector<nanogui::Button*> mCurrentImageButtons;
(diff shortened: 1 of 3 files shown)


Tom94:
Thanks, that really helps!

> Although in debug mode, when I open the image on my secondary monitor, NanoGUI throws a lot of complaints in the following format (but the monitor and window resize seem to be correct):

Pretty sure that's a bug in nanogui's (relatively) new size cache feature. I'll try fixing this separately. Definitely unrelated to this PR, but I appreciate you reporting it nonetheless!

> Another small note: on my primary monitor, when I load and then refresh the image, it shifts by one pixel. Looking into this more, somehow set_size() seems to be calling the correct size of (1766, 1000) when I load the image, but somehow m_size gets set to (1765, 1000) after that call finishes. In the function itself, it seems like this happens after glfwSetWindowSize(), but I don't see how this could possibly change the value of m_size unless some super weird memory accesses are happening

nanogui internally stores sizes that have the monitor's content scale factored in as integers, so necessarily does some rounding. `m_size` gets updated in a GLFW callback that reports successful resize of the window, and it ultimately gets set to `1765 = int(int(1766/1.25)*1.25)` where the division happened in `set_size`. It's indeed Windows (and, as it happens, also X11) specific -- monitor scale is a bit of a mess that has to be treated slightly differently on each of the OSs.

Anyway, you probably don't care about all that detail. Turns out if you add a round-up operation at the right place, you can ensure the integer that goes in is the same one that comes out, at least. I pushed a fix and would appreciate if you could try!

Better would be to change nanogui to use floating-point positioning throughout, but that'd be a huge change. 😅

> As a side note, is there a way to "lock" the window size when loading / refreshing images? It seems like right now, if the window size is smaller than the image, it will always resize to fit the image. It feels like there are some use cases where you may want the window to occupy a fixed amount of space (e.g., if I want tev to occupy half the monitor, while using the other half for something else). But in any case, I don't think this would be a high-priority issue.

Fair point. Quickly added a UI toggle and a CLI option (`--no-resize-window`).

PR #428, opened 2026-03-08
feat(image): color channel mask

PedroVillegas:
Hi Thomas,

I recently found out about tev and have found it to be very handy. The only thing I felt was missing: a color channel masking feature which lets me isolate individual, or multiple, color channels.

Note: I don't have an apple device to make and test changes for the metal shader, so I didn't.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ _CPack_Packages
 /.direnv
 /.vscode
 /.cache
+/.idea
 
 # Python cache
 __pycache__
--- a/include/tev/Common.h
+++ b/include/tev/Common.h
@@ -697,6 +697,72 @@ inline nanogui::Vector3f applyTonemap(nanogui::Vector3f value, float gamma, ETon
     return min(max(result, nanogui::Vector3f{0.0f}), nanogui::Vector3f{1.0f});
 }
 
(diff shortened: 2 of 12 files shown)


Tom94:
Hi Pedro, thanks for the PR!

tev can already explode the channels of the currently selected group into individual ones via Ctrl+X. This'll work even when the group isn't made up of RGBA like this PR assumes it to be. But... that feature is not very discoverable; just an entry in the keybindings help window. I assume you didn't see it before working on this PR?

Viewing group *subsets* is also possible, although admittedly quite a bit more cumbersome: use e.g. `tev :R,G image1.png image2.png ...` to load just the red and green channels of your images.

So, with these features in mind, I feel like a better approach to the PR would be to
- Add a GUI button for doing what Ctrl+X currently does, and
- (optionally) make it work with subset selection, e.g. using a GUI similar to what the current PR has, but generalized to the current channel group rather than hardcoded for RGBA. Also perhaps worth adding a toggle to make it apply to all currently-open images with the same channel group.
- Possibly both in a popup to avoid clutter. Especially since the current channel group can have arbitrarily many channels in it.

Compared to a pure shader-level approach, that'll also ensure that the histogram tracks the masked channels & that saving via Ctrl+S stays faithful to what's shown.

PR #436, opened 2026-04-03
feat(fits): add FITS image support

cek:


Tom94:
Heya, thanks a lot for the PR. Currently don't have all that much bandwidth, so might be a little bit until I get to review & merge.

cek:
No worries at all -- no rush, and thanks! I'm just happy to be able to examine astrophoto images in my favorite viewer, rather than having to use one of the terrible alternatives. ;)

To be a more full-featured replacement for said alternatives, tev would also need a astrophoto-centric tone mapping operator and associated controls.  However, such a specific/extreme tone mapper probably won't be of interest to most tev users, and thus would be a waste of UI space to render it all of the time. Once I have something working, I'm likely to have questions about how to integrate it such that it doesn't clutter the UI when not needed (e.g., a visibility toggle, defaulting to rendering it if a FITS image is loaded, for example...?) Or it may prove to be too specialized to add to tev proper. In any case, more to come.

Tom94:
I was planning to add soft clipping (log-space sigmoid with configurable min/max brightness) to the HDR popup menu. Probably not the kind of tone mapping you are looking for 😅, but mentioning it just in case.

I think other specialized tone maps would also fit into the HDR popup for now… to be consolidated in an actual tone map selector when/if tev gets support for a bunch of other standard graphics ones (that’d share aforementioned min/max brightness controls and also consume image metadata like max FALL etc.)

src/imageio/FitsImageLoader.cpp
```diff
@@ -0,0 +1,116 @@
...
+#define TINYFITS_IMPLEMENTATION
+#include <tinyfits/tinyfits.h>
+
+using namespace nanogui;
+using namespace std;
+
+namespace tev {
+
+Task<vector<ImageData>> FitsImageLoader::load(istream& iStream, const fs::path&, string_view, const ImageLoaderSettings&, int priority) const {
+    char magic[9];
```
Tom94: Should get zero-initialized via `= {};` to avoid UB.

Following two lines should use `sizeof(magic)` instead of repeating the hardcoded size.

src/imageio/FitsImageLoader.cpp
```diff
@@ -0,0 +1,116 @@
...
+    iStream.seekg(0, iStream.beg);
+
+    HeapArray<char> data(dataSize);
+    iStream.read(data.data(), dataSize);
+
+    TinyFits fits = {};
+    void* pixels = nullptr;
+    int err = tinyfits_load_from_memory(&fits, data.data(), (size_t)dataSize, &pixels);
+    if (err != TINYFITS_OK) {
+        tinyfits_free(&fits);
```
Tom94: All the explicit tinyfits_free and tinyfits_free_buffer calls should be replaced with a `ScopeGuard`. (See e.g. STBI image loader for a example usage.)

src/imageio/FitsImageLoader.cpp
```diff
@@ -0,0 +1,116 @@
...
+        throw ImageLoadError{"Image has zero pixels."};
+    }
+
+    const auto numChannels = (size_t)fits.num_channels;
+
+    // Convert native pixel data to normalized float32 for display.
+    // tinyfits_to_float normalizes unsigned integer types to [0,1]; floats pass through as-is.
+    const auto totalSamples = numPixels * numChannels;
+    vector<float> floatPixels(totalSamples);
```
Tom94: Better to create the channels upfront and then populate them directly via a switch over channel format w/
```cpp
co_await toFloat32((pixel_t*)pixels + numPixels * c, 1, channels[c].view(), false, priority);
```

Avoids the memcpy, extra allocation, and uses tev's thread pool to parallelize.

src/imageio/FitsImageLoader.cpp
```diff
@@ -0,0 +1,116 @@
...
+        resultData.channels.emplace_back(name, size, EPixelFormat::F32, EPixelFormat::F32);
+    }
+
+    // tinyfits_to_float returns planar data (all of channel 0, then channel 1, ...)
+    for (size_t c = 0; c < numChannels; ++c) {
+        auto view = resultData.channels[c].view<float>();
+        memcpy(view.data(), floatPixels.data() + c * numPixels, numPixels * sizeof(float));
+    }
+
+    resultData.hasPremultipliedAlpha = false;
```
Tom94: From the channel naming scheme above, it seems to me like FITS images can't contain an alpha channel at all? In that case, it'd be better to specify `resultData.hasPremultipliedAlpha = true` to avoid some tev-internal conversion checks.

src/imageio/FitsImageLoader.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+    AttributeNode root;
+    root.name = "FITS header";
+    AttributeNode& section = root.children.emplace_back();
+    section.name = "";
+    for (int i = 0; i < fits.num_keywords; ++i) {
+        AttributeNode node;
+        node.name = fits.keywords[i].key;
+        node.value = fits.keywords[i].value;
+        section.children.push_back(std::move(node));
```
Tom94: Rather than `std::move`, prefer direct emplace `AttributeNode& node = section.children.emplace_back();`

Similarly for the root node.

Tom94 (changes requested):
Thanks again for the PR! Looks mostly good to me -- just requesting a few minor changes.

EDIT: oh, and I almost forgot: could you add a fits entry to `void ImageViewer::openImageDialog()` as well?

src/imageio/FitsImageLoader.cpp
cek: Done

src/imageio/FitsImageLoader.cpp
cek: Nice; done.

src/imageio/FitsImageLoader.cpp
cek: Done.

src/imageio/FitsImageLoader.cpp
cek: Indeed; done.

src/imageio/FitsImageLoader.cpp
cek: Makes sense; done.

cek:
Thanks and sorry about that -- a little too quick on the PR trigger...

Tom94:
Thanks a lot! I think I can take it from here. Just one more question to make sure: are uint8 fits images really linearly encoded?

If they were sRGB-encoded you could use `toFloat32<true>(...)` to do the decoding, but generic transfers like gamma curves require a bit of extra code.

And: is there no signed int8 format?

cek:
The vast majority of FITS files hold either linear 12-to-16-bit raw sensor data as uint16, or linear processed (and possibly negative) data as fp32.

They can also hold processed data that is wildly nonlinear -- extreme "stretching" of shadow/lower-order bits where the interesting astro signals live. But FITS data is essentially never sRGB-encoded.

FITS has no real support for signed 8-bit data. I don't think I've ever seen an 8-bit FITS file in the wild, to be honest.

Thanks again!

Tom94:
Sorry again for the delay; just tested the PR and it's ready to be merged in principle. Just one remaining thing:

I noticed that 5 out of the 10 top-level sample images from https://fits.gsfc.nasa.gov/fits_samples.html don't load. 4 of them with "Invalid FITS structure" and 1 with "Non-standard BZERO/BSCALE".

I think it's fine if tinyfits does not support the whole format (though I'd of course prefer that :)), but the error messages should at least make clear which FITS images are simply unsupported vs. malformed. **tev**'s README should also specify which subset of FITS is actually supported in the formats section. (**tev** goes to great lengths to support pretty much every feature of the existing bolded formats.)

cek:
Well, you shamed me into doing what I should have originally done -- make `tinyfits`  less of a hack, and support a more complete subset of the FITS standard, and supply more helpful error messages. Thanks! ;)

Support is still very much a subset of full FITS functionality, however. The FITS standard dates back to the 1970s -- header data fields were originally limited to punchcard-compatible lengths!  FITS files can store a wide variety (and arbitrary count) of "header-data units" (HDUs), not  just single images. Two of the FITS files you referenced above don't actually contain images, for example, and one contains multiple images.

The FITS standard has grown over the years, and the reference FITS read/write library, cfitsio, consists of more than 150k LOC.

To limit scope, `tinyfits` is primarily focused on the subset of FITS functionality used in amateur astrophotography. In a nutshell, this means that it will only load data from the first "primary" HDU or "image extension" HDU that holds 2- or 3-dimensional data.  Data tables, spectra, multiple images, compressed images, "Random Groups," etc. are supported by FITS, but are not used in practice by amateur astrophoto hardware or software, and can't be loaded by `tinyfits`.

In any case, I've updated the PR to use the new version of `tinyfits` and to provide more helpful error messages.

I've made one change that I'm not crazy about: FITS header fields often comes with a comment string that is useful to display in the image's information dialog. To do so, the PR hijacks the dialog's "type" column to display comments. The resulting layout isn't great (comments tend to be longer than unit names), but I didn't want to mess with the dialog layout or size without a discussion.

I still need to update the README as you requested.  But before I do, I'm now thinking that full-blown FITS support may be too niche for most tev users. Displaying raw FITS images in tev is a big quality-of-life improvement for me. But a huge jump in utility would come from also adding debayering and astrophoto-specific tonemapping options, which are unlikely to be of interest to most tev users.

Perhaps this work is better suited for a branch, for an astrophoto-focused fork, or just my own private fork. ;)  Or perhaps basic FITS reading is useful enough to land in tev proper, but the astrophoto-specific display options should land elsewhere.

What do you think? I'm genuinely happy to go in whatever direction seems best to you -- I'm just happy to have a really good FITS viewer!

Tom94:
Thanks a bunch for the update & changes!

> I've made one change that I'm not crazy about: FITS header fields often comes with a comment string that is useful to display in the image's information dialog. To do so, the PR hijacks the dialog's "type" column to display comments. The resulting layout isn't great (comments tend to be longer than unit names), but I didn't want to mess with the dialog layout or size without a discussion.

No qualms with the hijacking part at all. To fix layouting, I'm thinking a "comment" column in the image info window might make sense in general, but that's best left to a separate PR, I think.

> But a huge jump in utility would come from also adding debayering and astrophoto-specific tonemapping options, which are unlikely to be of interest to most tev users.

Is `Demosaic.h/.cpp` useful for this at all? It implements the AMaZE debayering algorithm for the standard 1r2g1b bayer pattern, as well as a general, lower-quality algorithm for arbitrary cfa patterns.

It's currently only used when reading DNG images (via `TiffImageLoader`) where demosaicing is performed by default and the raw image is kept as an additional channel called `cfa.L`. The FITS loader could do the same.

As for tonemapping: if interactive controls aren't important, it could be applied in `FitsImageLoader` and controlled via the `ImageLoaderSettings` struct that's passed to `load(...)`. Analogously to how gainmap and DNG camera profile application is currently being configured.

Otherwise, of course no expectations from my end for you to perform any of that work. Or to not just fork tev and hack in whatever things you would like with a lot less effort than required by upstreaming. :p I'm already happy you're contributing this PR in the first place and will be happy to merge even if you leave it as is (minus the small README addendum).

> The FITS standard has grown over the years, and the reference FITS read/write library, cfitsio, consists of more than 150k LOC.

I wouldn't actually mind depending on `cfitsio` if it makes integration easier. I suspect that even 150k LoC of C code wouldn't significantly affect tev's compilation time... most other supported formats already use heavy-duty reference implementations in favor of tiny implementations (jpg, png, avif, heic, jxl, etc.). But then again, really not expecting you to perform that work -- though I might give the switch a shot at some point in the future. Features like multi-image support and data coordinate specifications *would* mesh well with some of tev's existing EXR features. :)

EDIT: on the other hand -- and I don't know how difficult this would be to add -- if tinyfits supported all the image-related features of FITS and exposed access to compressed pixel data, even if it can't decompress it by itself, there'd be no reason to ever switch to cfitsio. Many of tev's image loaders already handle exotic compression schemes internally.

cek:
If you don't have an issue with adopting cfitsio, tev would almost certainly be better served relying on a codebase that will outlive all of us, than on my side-project of a side-project.

I can close this PR and either let you have a crack at cfitsio if you're keen to do so, or I can try putting together a new PR based on it.

Interactive tonemapping controls would be extremely useful, but it I imagine it would be easy to make them appear on demand rather than cluttering the core UI.

I didn't realize there was already an AMaZE implementation in the codebase. That will do just fine, I'm sure.

Tom94:
> I can close this PR and either let you have a crack at cfitsio if you're keen to do so, or I can try putting together a new PR based on it.

I'm more than happy to let you do it if you're up for it! And no need to close the PR -- if you give me the go ahead I'm happy to merge as-is to get basic FITS support going until moving to cfitsio at some point in the future.

> Interactive tonemapping controls would be extremely useful, but it I imagine it would be easy to make them appear on demand rather than cluttering the core UI.

That's very fair. tev could use a proper tonemapping submenu anyway, not just with a bunch of astrophotography operators, but also standard graphics ones (thinking of the typical suspects like ACES, Hable, Reinhard, ...), but those can come later. Please feel free to PR the astrophotography operators you need at any time.

In terms of GUI, I think an "Advanced" popup button to the right of the current tonemapping options would work well, like this:

image

PR #442, opened 2026-05-05
feat(fits): add fits image read support

cek:


Tom94:
Oh wow, that was fast. Thanks a bunch!

On a quick test, I have to enable `-DUSE_PTHREADS=ON` to not run into race conditions when loading a folder full of test images. A bit annoying, because it'll need https://github.com/jwinarske/pthreads4w looped in just for MSVC, but eh, it's a small enough dependency and the cmake tame enough to `add_subdirectory(pthreads4w EXCLUDE_FROM_ALL)`, I reckon.

(Alternatively, we could fork cfitsio and implement Windows-native mutexes behind some ifdefs. On a quick `grep`, pthreads isn't actually used in many places and relatively well isolated. But still seems more trouble than going with pthreads4w. :p)

cek:
Ah, yes, I should have remembered that cfitsio maintains a pile of global state. I did stress test loading using a couple folders full of images, but it seems I just got lucky.

It's a relatively small change to give `cfitsio` the option of using C11's `<threads.h>` instead of pthreads -- see [here](https://github.com/HEASARC/cfitsio/compare/develop...cek:cfitsio-tev:c11-threads). But of course forking has costs. Maybe I should see if the cfitsio folks are interested in an MR...

In any case, I'll leave it to you to decide which is the lesser evil:  forking `cfitsio`,  or relying on `pthreads4w`.

Tom94:
Nice, C11 threads are even better! Since you already got them working, let's just go with the fork. That'll also let us get rid of the `unistd` dance and `ExternalProject_Add`, which I'd prefer avoiding if possible. (Existing uses of `ExternalProject_Add` are projects with much more complex-to-disentangle CMake-side issues, sadly.)

Half of tev's dependencies already are forks because of various CMake idiosyncrasies (and in some cases similar fixups like the pthreads one here), so no harm maintaining another -- especially with such a small patch on top of upstream. Heck, tev's linux HDR support is a *huge* divergence from upstream GLFW. 😅

cek:
I've updated the MR to use a [fork](https://github.com/cek/cfitsio-tev/tree/tev) of `cfitsio` that includes `CMakeLists.txt` changes that allow it to be included as a subproject via `add_subdirectory()`, removes the `unistd.h` issue, and adds C11 threads as an option (enabled on WIN32 only after realizing macOS support was only added quite recently).

You'll obviously want your own cfitsio fork if this lands.

I'll see if the cfitsio project is interested in cmake or C11 threads MRs, but I suspect that's a long shot.

